PR #10, opened 2024-03-21
Add Current Test/Class elements

FriggaHel:
# One-line summary

> Issue : IRIS-834

## Description

- Exposes fields to set context
- Move classes from `public` to `internal` as they should not be exposed
- Expose `CurrentTestName` & `CurrentClassName` so it can be overridden by user.

## Types of Changes

### DIFF
--- a/visual-dotnet/SauceLabs.Visual/VisualCheckOptions.cs
+++ b/visual-dotnet/SauceLabs.Visual/VisualCheckOptions.cs
@@ -15,7 +15,13 @@ namespace SauceLabs.Visual
         public bool? CaptureDom { get; set; }
         public string? ClipSelector { get; set; }
 
+        /// <summary>
+        /// <c>SuiteName</c> manually set the SuiteName of the Test.
+        /// </summary>
         public string? SuiteName { get; set; }
+        /// <summary>
+        /// <c>TestName</c> manually set the TestName of the Test.
+        /// </summary>
         public string? TestName { get; set; }
     }
 }
--- a/visual-dotnet/SauceLabs.Visual/VisualCheckOptions.cs
+++ b/visual-dotnet/SauceLabs.Visual/VisualCheckOptions.cs
@@ -28,15 +28,17 @@ namespace SauceLabs.Visual
 
         private bool HasIncompleteTestContext() => string.IsNullOrEmpty(SuiteName) || string.IsNullOrEmpty(TestName);
 
-        internal void PopulateTestContext(string callerMemberName, string? previousSuiteName)
+        internal void EnsureTestContextIsPopulated(string callerMemberName, string? previousSuiteName)
         {
-            if (!string.IsNullOrEmpty(callerMemberName) && HasIncompleteTestContext())
+            if (string.IsNullOrEmpty(callerMemberName) || HasIncompleteTestContext())
             {
-                var stack = new StackTrace();
-                var frame = stack.GetFrames()?.FirstOrDefault(f => f.GetMethod().Name == callerMemberName);
-                SuiteName ??= frame?.GetMethod().DeclaringType?.FullName ?? previousSuiteName;
-                TestName ??= callerMemberName;
+                return;
             }
+
+            var stack = new StackTrace();
+            var frame = stack.GetFrames()?.FirstOrDefault(f => f.GetMethod().Name == callerMemberName);
+            SuiteName ??= frame?.GetMethod().DeclaringType?.FullName ?? previousSuiteName;
+            TestName ??= callerMemberName;
         }
     }
 }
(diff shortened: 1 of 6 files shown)


visual-dotnet/SauceLabs.Visual/VisualCheckOptions.cs
```diff
@@ -13,5 +13,8 @@ public class VisualCheckOptions
         public IgnoreRegion[]? IgnoreRegions { get; set; }
         public IWebElement[]? IgnoreElements { get; set; }
         public bool? CaptureDom { get; set; }
+
+        public string? ClassName { get; set; }
```
konraddysput, line 17: How about following the convention from webdriver?
```suggestion
        public string? SuiteName { get; set; }
```

visual-dotnet/SauceLabs.Visual/VisualClient.cs
```diff
@@ -26,6 +29,12 @@ public class VisualClient : IDisposable
         public bool CaptureDom { get; set; } = false;
         private readonly ResiliencePipeline _retryPipeline;
 
+        public string? CurrentTestClass { get; set; }
+        public string? CurrentTestName { get; set; }
```
konraddysput, line 33: If we move to the singleton instnace + we will run tests in parallel - it would be super hard to keep this API working correctly. 

We also don't have any reference to it - I think we should remove them, right?

visual-dotnet/SauceLabs.Visual/VisualClient.cs
```diff
@@ -26,6 +29,12 @@ public class VisualClient : IDisposable
...
+        private string? _previousTestClass = null;
+
+        #region Create
```
konraddysput, line 36: smal note aboute regions:

They're considered as anti-pattern because if you want to use them it means your class it to big. 

I'm ok but maybe we should consider moving some business logic from this to small testable classes?

visual-dotnet/SauceLabs.Visual/VisualClient.cs
```diff
@@ -193,17 +205,21 @@ private async Task<VisualBuild> FindBuildById(string buildId)
         /// <returns>a <c>VisualBuild</c> instance</returns>
         private async Task<VisualBuild> CreateBuild(CreateBuildOptions? options = null)
         {
+            var projectName = options?.Project ?? Assembly.GetExecutingAssembly().FullName;
```
konraddysput, line 208: Base on what we saw I think its safer to use "Unknown" rather than executing assembly - wdyt?

konraddysput (changes requested):

visual-dotnet/SauceLabs.Visual/VisualClient.cs
FriggaHel: Let's remove it 👍  It was mainly allowing to group collapsing parts.

visual-dotnet/SauceLabs.Visual/VisualClient.cs
FriggaHel: We should remove it, and probably keep default behaviour.

visual-dotnet/SauceLabs.Visual/VisualClient.cs
```diff
@@ -220,22 +225,39 @@ private async Task FinishBuild(VisualBuild build)
         /// <param name="name">the name of the screenshot</param>
         /// <param name="options">the configuration for the screenshot capture and comparison</param>
         /// <returns></returns>
-        public async Task<string> VisualCheck(string name, VisualCheckOptions? options = null)
+        public Task<string> VisualCheck(string name, VisualCheckOptions? options = null,
+            [CallerMemberName] string callerMemberName = "")
+        {
+            options ??= new VisualCheckOptions();
+            if (!string.IsNullOrEmpty(callerMemberName) && (string.IsNullOrEmpty(options.SuiteName) || string.IsNullOrEmpty(options.TestName)))
```
konraddysput, line 232: small nit:
maybe we can keep the business logic here clean and expose it to a method or util?

visual-dotnet/SauceLabs.Visual/VisualCheckOptions.cs
```diff
@@ -14,5 +14,8 @@ public class VisualCheckOptions
         public IWebElement[]? IgnoreElements { get; set; }
         public bool? CaptureDom { get; set; }
         public string? ClipSelector { get; set; }
+
+        public string? SuiteName { get; set; }
```
konraddysput, line 18: should we provide a comment for public interface properties?

konraddysput (approved):

konraddysput (approved):

konraddysput (approved):

PR #11, opened 2024-03-21
Add support for clipSelector

FriggaHel:
# One-line summary

> Issue : IRIS-855

## Description

Add support for `clipSelector`.

## Types of Changes

### DIFF
--- a/visual-dotnet/SauceLabs.Visual/GraphQL/CreateSnapshotFromWebDriverIn.cs
+++ b/visual-dotnet/SauceLabs.Visual/GraphQL/CreateSnapshotFromWebDriverIn.cs
@@ -26,6 +26,9 @@ namespace SauceLabs.Visual.GraphQL
         [JsonProperty("captureDom")]
         public bool? CaptureDom { get; set; }
 
+        [JsonProperty("clipSelector")]
+        public string? ClipSelector { get; set; }
+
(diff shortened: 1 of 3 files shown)


visual-dotnet/SauceLabs.Visual/GraphQL/CreateSnapshotFromWebDriverIn.cs
```diff
@@ -26,6 +26,9 @@ internal class CreateSnapshotFromWebDriverIn
         [JsonProperty("captureDom")]
         public bool? CaptureDom { get; set; }
 
+        [JsonProperty("clipSelector")]
+        public string? ClipSelector { get; set; }
+
         public CreateSnapshotFromWebDriverIn(
```
konraddysput, line 32: I think we could just leave the object as it is and require the user to set properties via setters. The constructor is getting bigger.

konraddysput (approved):

visual-dotnet/SauceLabs.Visual/GraphQL/CreateSnapshotFromWebDriverIn.cs
FriggaHel: This constructor is not used by the customer so we can modify it as we please.
I'll simplify it in another MR to rely mainly on properties.

PR #12, opened 2024-03-25
Ensure build is shared across all test classes

FriggaHel:
# One-line summary

> Issue : IRIS-856

## Description

Currently build is created (or loaded) upon creation of the `VisualClient`.
This PR comes to persist the newly created build, thus allowing multiple class to be reported within the same `VisualBuild`.

As a consequence, the `Cleanup` will need to be moved out of the test AfterAll, but to a global AfterAll.

Additionally:
- Fixes a bug in the `TextContext` implementation
- Removes dependency of `VisualApi` to `WebDriver` as it's never used

## Types of Changes

### DIFF
--- a/visual-dotnet/SauceLabs.Visual/BuildFactory.cs
+++ b/visual-dotnet/SauceLabs.Visual/BuildFactory.cs
@@ -1,4 +1,3 @@
-using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
--- a/visual-dotnet/SauceLabs.Visual/BuildFactory.cs
+++ b/visual-dotnet/SauceLabs.Visual/BuildFactory.cs
@@ -8,24 +8,24 @@ namespace SauceLabs.Visual
 {
     internal static class BuildFactory
     {
-        private static readonly Dictionary<Region, VisualBuild> _builds = new Dictionary<Region, VisualBuild>();
+        private static readonly Dictionary<string, VisualBuild> _builds = new Dictionary<string, VisualBuild>();
 
         internal static async Task<VisualBuild> Get(VisualClient client, CreateBuildOptions options)
         {
             // Check if there is already a build for the current region.
-            if (_builds.TryGetValue(client.Api.Region, out var build))
+            if (_builds.TryGetValue(client.Api.Region.Name, out var build))
             {
                 return build;
             }
 
             var createdBuild = await Create(client, options);
-            _builds[client.Api.Region] = createdBuild;
+            _builds[client.Api.Region.Name] = createdBuild;
             return createdBuild;
         }
 
         private static void Disregard(VisualBuild build)
         {
-            Region? key = null;
+            string? key = null;
             var enumerator = _builds.GetEnumerator();
             while (enumerator.MoveNext())
             {
--- a/visual-dotnet/SauceLabs.Visual/BuildFactory.cs
+++ b/visual-dotnet/SauceLabs.Visual/BuildFactory.cs
@@ -29,19 +29,7 @@ namespace SauceLabs.Visual
         /// <param name="build">the build to remove</param>
         private static void Disregard(VisualBuild build)
         {
-            string? key = null;
-            var enumerator = Builds.GetEnumerator();
-            while (enumerator.MoveNext())
-            {
-                if (enumerator.Current.Value != build)
-                {
-                    continue;
-                }
-                key = enumerator.Current.Key;
-                break;
-            }
-            enumerator.Dispose();
-
+            var key = (from entry in Builds where entry.Value == build select entry.Key).FirstOrDefault();
             if (key != null)
             {
                 Builds.Remove(key);
--- a/visual-dotnet/SauceLabs.Visual/BuildFactory.cs
+++ b/visual-dotnet/SauceLabs.Visual/BuildFactory.cs
@@ -50,13 +50,37 @@ namespace SauceLabs.Visual
             var key = FindRegionByBuild(build);
             if (key != null)
             {
-                var entry = Builds[key];
+                await Close(key, Builds[key]);
+            }
+        }
+
+        /// <summary>
+        /// <c>Close</c> finishes and forget about <c>build</c>
+        /// </summary>
+        /// <param name="region">the build to finish</param>
+        /// <param name="entry">the api/build pair</param>
+        private static async Task Close(string region, ApiBuildPair entry)
+        {
+            if (!entry.Build.IsExternal)
+            {
                 if (!entry.Build.IsExternal)
                 {
                     await entry.Api.FinishBuild(entry.Build.Id);
                 }
+            }
+            Builds.Remove(region);
+            entry.Api.Dispose();
+        }
 
-                Builds.Remove(key);
+        /// <summary>
+        /// <c>CloseBuilds</c> closes all build that are still open.
+        /// </summary>
+        internal static async Task CloseBuilds()
+        {
+            var regions = Builds.Keys;
+            foreach (var region in regions)
+            {
+                await Close(region, Builds[region]);
             }
         }
 
@@ -97,24 +121,6 @@ namespace SauceLabs.Visual
             catch (VisualClientException)
             {
                 return null;
-            }
-        }
-
-        /// <summary>
-        /// <c>CloseBuilds</c> closes all build that are still open.
-        /// </summary>
-        internal static async Task CloseBuilds()
-        {
-            var regions = Builds.Keys.ToArray();
-            foreach (var region in regions)
-            {
-                var entry = Builds[region];
-                if (!entry.Build.IsExternal)
-                {
-                    await entry.Api.FinishBuild(entry.Build.Id);
-                }
-                entry.Api.Dispose();
-                Builds.Remove(region);
             }
         }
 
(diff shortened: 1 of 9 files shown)


visual-dotnet/SauceLabs.Visual/VisualClient.cs
```diff
@@ -91,15 +99,24 @@ private async Task SetupBuild(CreateBuildOptions buildOptions)
...
+            Build = new VisualBuild(createBuildResponse.Id, createBuildResponse.Url, createBuildResponse.Mode)
             {
-                buildOptions.CustomId ??= EnvVars.CustomId;
-                var createBuildResponse = await CreateBuild(buildOptions);
-                Build = new VisualBuild(createBuildResponse.Id, createBuildResponse.Url, createBuildResponse.Mode);
-                _externalBuild = false;
-            }
+                IsExternal = false
+            };
+            var copiedApi = _api.Clone();
```
konraddysput, line 113: based on the design we're allowing to send data to a different instances (based on region credentials etc). We're giving this impression to our customers based on our API design. In the end it doesn't matter because the API is also cloned, so their changes are not respected.

How about leaving the API client dedicated for each visual client?

visual-dotnet/SauceLabs.Visual/VisualClient.cs
```diff
@@ -79,10 +81,16 @@ public static async Task<VisualClient> Create(WebDriver wd, Region region, strin
 
         private async Task SetupBuild(CreateBuildOptions buildOptions)
         {
-            var response = await this._api.WebDriverSessionInfo(_jobId, _sessionId);
+            var response = await _api.WebDriverSessionInfo(_jobId, _sessionId);
             var metadata = response.EnsureValidResponse();
             _sessionMetadataBlob = metadata.Result.Blob;
 
+            if (_sharedBuild != null)
```
konraddysput, line 88: I think we should hide what exactly is under the shared build by using some kind of abstraction - why? I think we should keep the client code as clean as possible.


How about moving this code to BuildFactory responsible for generating/managing build?

visual-dotnet/SauceLabs.Visual/VisualClient.cs
FriggaHel: Leaving the API client dedicated for each visual will generate probably more issues.
We expect users to dispose of `VisualClient` after the class has been processed. That will dispose both the `VisualApi` instance and it's internal `GraphQLClient`. Cloning the `VisualApi` itself, ensure that it it still valid when it needs to be invoked.

visual-dotnet/SauceLabs.Visual/VisualClient.cs
FriggaHel: You're right. This will also allow to have a build that is shared but that could still be isolated by region (which is not the case yet).

In a typical usecase that should not happen, but that would be a flaw.

visual-dotnet/SauceLabs.Visual/BuildFactory.cs
```diff
@@ -0,0 +1,172 @@
...
+                DefaultBranch = options.DefaultBranch,
+            })).EnsureValidResponse();
+
+            build = new VisualBuild(result.Result.Id, result.Result.Url, result.Result.Mode)
+            {
+                IsExternal = false
+            };
+
+            var copiedApi = api.Clone();
+            build.Close = async () =>
```
konraddysput, line 143: maybe it's a microoptimization but maybe from the code perspective it would look better. How about not only caching the build but also an HTTP client?
then we don't need to even make a copy of the client. Every time you make a copy you also reserve a port and other resources. Maybe it's worth to also cache it?

visual-dotnet/SauceLabs.Visual/VisualBuild.cs
```diff
@@ -12,6 +14,9 @@ public class VisualBuild
 
         public BuildMode Mode { get; internal set; }
 
+        internal bool IsExternal { get; set; } = false;
```
konraddysput, line 17: can we add a comment what does it mean?

visual-dotnet/SauceLabs.Visual/VisualCheckOptions.cs
```diff
@@ -30,7 +30,7 @@ public class VisualCheckOptions
 
         internal void EnsureTestContextIsPopulated(string callerMemberName, string? previousSuiteName)
         {
-            if (string.IsNullOrEmpty(callerMemberName) || HasIncompleteTestContext())
+            if (string.IsNullOrEmpty(callerMemberName) || !HasIncompleteTestContext())
```
konraddysput, line 33: is this correct?

visual-dotnet/SauceLabs.Visual/VisualClient.cs
```diff
@@ -256,19 +163,27 @@ private async Task<string> VisualCheckAsync(string name, VisualCheckOptions opti
         }
 
         /// <summary>
-        /// <c>Cleanup</c> set a correct status to the build. No action should be made after that calling <c>Cleanup</c>.
+        /// <c>Cleanup</c> set a correct status to the build. No action should be made after calling <c>Cleanup</c>.
         /// </summary>
         public async Task Cleanup()
         {
-            if (!_externalBuild)
+            if (Build.Close != null)
```
konraddysput, line 170: should we raise an exception when the close method is undefined? Maybe it would be better to do it and fail hard rather than silently fail

visual-dotnet/SauceLabs.Visual/VisualCheckOptions.cs
FriggaHel: Yes, there is an actual bug in the main branch.  I can make another PR for that if needed.

visual-dotnet/SauceLabs.Visual/VisualClient.cs
FriggaHel: Removed

visual-dotnet/SauceLabs.Visual/BuildFactory.cs
FriggaHel: Correct me if I'm wrong, but until a request has been made, it should not reserve anything network related. A bit of memory yes, but not sure it uses a port at this stage.

visual-dotnet/SauceLabs.Visual/BuildFactory.cs
```diff
@@ -0,0 +1,172 @@
...
+            var build = await GetEffectiveBuild(api, EnvVars.BuildId, EnvVars.CustomId);
+            if (build != null)
+            {
+                if (!build.IsRunning())
+                {
+                    throw new VisualClientException($"build {build.Id} is not RUNNING");
+                }
+
+                build.IsExternal = true;
+                build.Close = async () => Disregard(build);
```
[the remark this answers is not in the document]
FriggaHel: 1. Compiler refused to do so.
2. True, will address that.

visual-dotnet/SauceLabs.Visual/BuildFactory.cs
```diff
@@ -0,0 +1,172 @@
...
+
+        /// <summary>
+        /// <c>Disregard</c> removes the build from the known builds.
+        /// </summary>
+        /// <param name="build">the build to remove</param>
+        private static void Disregard(VisualBuild build)
+        {
+            string? key = null;
+            var enumerator = Builds.GetEnumerator();
+            while (enumerator.MoveNext())
```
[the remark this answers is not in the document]
FriggaHel: Fixed

visual-dotnet/SauceLabs.Visual/BuildFactory.cs
```diff
@@ -8,40 +9,44 @@ namespace SauceLabs.Visual
...
 
             var createdBuild = await Create(api, options);
-            Builds[api.Region.Name] = createdBuild;
+            Builds[api.Region.Name] = new ApiBuildPair(api.Clone(), createdBuild);
             return createdBuild;
         }
 
+        private static string? FindRegionByBuild(VisualBuild build)
+        {
+            return (from entry in Builds where entry.Value.Build == build select entry.Key).FirstOrDefault();
```
konraddysput, line 41: I recommend not to use this LINQ syntax. 

instead please use 

`Build.GetValues().FirstOrDefault(n => n.Build == build)?.Key;`

visual-dotnet/SauceLabs.Visual/BuildFactory.cs
```diff
@@ -8,40 +9,44 @@ namespace SauceLabs.Visual
...
+
+        internal static async Task Close(VisualBuild build)
+        {
+            var key = FindRegionByBuild(build);
+            if (key != null)
+            {
+                var entry = Builds[key];
+                await entry.Api.FinishBuild(entry.Build.Id);
+                Builds.Remove(key);
+            }
```
konraddysput: one less indent and check via recommended method
```suggestion
            if (string.IsNullOrEmpty(key)
            {
                return;
            }
            var entry = Builds[key];
            await entry.Api.FinishBuild(entry.Build.Id);
            Builds.Remove(key);
```

visual-dotnet/SauceLabs.Visual/VisualCheckOptions.cs
```diff
@@ -26,11 +26,11 @@ public class VisualCheckOptions
         /// </summary>
         public string? TestName { get; set; }
 
-        private bool HasIncompleteTestContext() => string.IsNullOrEmpty(SuiteName) || string.IsNullOrEmpty(TestName);
+        private bool HasCompleteTestContext() => !string.IsNullOrEmpty(SuiteName) & !string.IsNullOrEmpty(TestName);
```
konraddysput, line 29: can we convert it into a regular method?

konraddysput (approved):

PR #14, opened 2024-04-02
.Net - Add support for Full Page Screenshot

paweltomaszewskisaucelabs:
## Description
Full page screenshot for c#

## Types of Changes

### DIFF
--- a/visual-dotnet/SauceLabs.Visual/FullPageConfig.cs
+++ b/visual-dotnet/SauceLabs.Visual/FullPageConfig.cs
@@ -1,0 +1,19 @@
+using SauceLabs.Visual.GraphQL;
+
+namespace SauceLabs.Visual
+{
+    public class FullPageConfig
+    {
+        public int? DelayAfterScrollMs { get; set; }
+        public string[]? HideAfterFirstScroll { get; set; }
+
+        internal FullPageConfigIn ToFullPageConfigIn()
+        {
+            return new FullPageConfigIn()
+            {
+                DelayAfterScrollMs = DelayAfterScrollMs,
+                HideAfterFirstScroll = HideAfterFirstScroll
+            };
+        }
+    }
+}
--- a/visual-dotnet/SauceLabs.Visual/FullPageConfig.cs
+++ b/visual-dotnet/SauceLabs.Visual/FullPageConfig.cs
@@ -1,3 +1,4 @@
+using System.Collections;
 using SauceLabs.Visual.GraphQL;
 
 namespace SauceLabs.Visual
@@ -5,7 +6,7 @@ namespace SauceLabs.Visual
     public class FullPageConfig
     {
         public int? DelayAfterScrollMs { get; set; }
-        public string[]? HideAfterFirstScroll { get; set; }
+        public IEnumerable? HideAfterFirstScroll { get; set; }
 
         internal FullPageConfigIn ToFullPageConfigIn()
         {
--- a/visual-dotnet/SauceLabs.Visual/FullPageConfig.cs
+++ b/visual-dotnet/SauceLabs.Visual/FullPageConfig.cs
@@ -1,4 +1,4 @@
-using System.Collections;
+using System.Collections.Generic;
 using SauceLabs.Visual.GraphQL;
 
 namespace SauceLabs.Visual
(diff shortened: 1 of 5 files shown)


visual-dotnet/SauceLabs.Visual/VisualCheckOptions.cs
```diff
@@ -11,10 +11,23 @@ namespace SauceLabs.Visual
...
         public DiffingMethod? DiffingMethod { get; set; }
         public IgnoreRegion[]? IgnoreRegions { get; set; }
         public IWebElement[]? IgnoreElements { get; set; }
         public bool? CaptureDom { get; set; }
+        public bool? FullPage { get; set; }
+        public FullPageConfig? FullPageConfig
+        {
+            get
+            {
+                return _fullPageConfig ?? (FullPage == true ? new FullPageConfig() : null);
```
FriggaHel, line 24: That logic should be in `VisualClient`.

`public FullPageConfig? FullPageConfig { get; set; }`

Then `VisualClient` will take care of dealing with that behaviour.

paweltomaszewskisaucelabs: Why it should be in `VisualClient`? If you will need to check if FullPage is turned on in other place you will duplicate the code.

paweltomaszewskisaucelabs: I also think that logic is related to `VisualCheckOptions`

visual-dotnet/SauceLabs.Visual/FullPageConfig.cs
```diff
@@ -0,0 +1,19 @@
+using SauceLabs.Visual.GraphQL;
+
+namespace SauceLabs.Visual
+{
+    public class FullPageConfig
```
konraddysput, line 5: Do we need to have two classes:
- FullPageConfigIn 
- FullpageConfig

if in the end we want to only use FullpageConfig and we know how to serialize it correctly?

visual-dotnet/SauceLabs.Visual/FullPageConfig.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public int? DelayAfterScrollMs { get; set; }
+        public string[]? HideAfterFirstScroll { get; set; }
```
konraddysput, line 8: Can we use IEnumerable<stirng>? instead

konraddysput (changes requested):
Can we please use IEnumerable instead?

visual-dotnet/SauceLabs.Visual/FullPageConfig.cs
FriggaHel: Idea was to avoid `.GraphQL` package to have a dependency to an higher package + not exposing raw content of `FullPageConfigIn` to user (since it does not provides the `Enable` flag natively.

visual-dotnet/SauceLabs.Visual/VisualClient.cs
```diff
@@ -144,6 +144,7 @@ private async Task<string> VisualCheckAsync(string name, VisualCheckOptions opti
             var ignored = new List<RegionIn>();
             ignored.AddRange(options.IgnoreRegions?.Select(r => new RegionIn(r)) ?? new List<RegionIn>());
             ignored.AddRange(options.IgnoreElements?.Select(r => new RegionIn(r)) ?? new List<RegionIn>());
+            var fullPageConfig = (options.FullPageConfig ?? new FullPageConfig()).ToFullPageConfigIn();
```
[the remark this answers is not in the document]
paweltomaszewskisaucelabs: It's fine to me. I have just adjusted the code.

visual-dotnet/SauceLabs.Visual/GraphQL/FullPageConfigIn.cs
```diff
@@ -7,6 +8,6 @@ internal class FullPageConfigIn
         [JsonProperty("delayAfterScrollMs")]
         public int? DelayAfterScrollMs { get; set; }
         [JsonProperty("hideAfterFirstScroll")]
-        public string[]? HideAfterFirstScroll { get; set; }
+        public IEnumerable? HideAfterFirstScroll { get; set; }
```
FriggaHel, line 11: I think that's preferable to use generics with `IEnumerable`, so we know the type that is in it.
```suggestion
        public IEnumerable<string>? HideAfterFirstScroll { get; set; }
```

konraddysput (approved):

FriggaHel (approved):

PR #15, opened 2024-04-03
Python & Robot Framework Support

omacranger:
# One-line summary

> Issue : #1234 (only if appropriate)

## Description
Adds Python & Robot Framework support.

## Types of Changes

## Tasks
_List of tasks you will do to complete the PR_
  - [ ] Created Task 1
  - [ ] Created Task 2
  - [ ] To-do Task 3

## Review
_List of tasks the reviewer must do to review the PR_
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG


visual-python/src/saucelabs_visual/client.py
```diff
@@ -0,0 +1,185 @@
...
+    def __init__(self):
+        self._create_client()
+
+    def _create_client(self):
+        region_url = Region.from_name(environ.get("SAUCE_REGION") or 'us-west-1').graphql_endpoint
+        username = environ.get("SAUCE_USERNAME")
+        access_key = environ.get("SAUCE_ACCESS_KEY")
+
+        if username is None or access_key is None:
+            raise Exception(
```
konraddysput, line 24: should we include environment variable names in the exception message?

visual-python/src/saucelabs_visual/client.py
```diff
@@ -0,0 +1,185 @@
...
+                'Sauce Labs credentials not set. Please check that you set correctly your `user` '
+                'and `key`.'
+            )
+
+        transport = AIOHTTPTransport(
+            url=region_url,
+            auth=BasicAuth(environ.get("SAUCE_USERNAME"), environ.get("SAUCE_ACCESS_KEY")),
```
konraddysput, line 31: you already have username and access key - line 20 & 21
```suggestion
            auth=BasicAuth(username, access_key),
```

visual-python/src/saucelabs_visual/client.py
```diff
@@ -0,0 +1,185 @@
...
+
+class SauceLabsVisual:
+    client: Client = None
+    build_id: str = None
+
+    def __init__(self):
+        self._create_client()
+
+    def _create_client(self):
+        region_url = Region.from_name(environ.get("SAUCE_REGION") or 'us-west-1').graphql_endpoint
```
konraddysput, line 19: small nit: 
if username or access key is not defined, there is no reason to read and create region URL. This code can be executed after the if statement from L23

visual-python/src/saucelabs_visual/client.py
```diff
@@ -0,0 +1,185 @@
...
+        transport = AIOHTTPTransport(
+            url=region_url,
+            auth=BasicAuth(environ.get("SAUCE_USERNAME"), environ.get("SAUCE_ACCESS_KEY")),
+        )
+        self.client = Client(transport=transport)
+
+    def get_client(self) -> Client:
+        return self.client
+
+    def create_build(
```
konraddysput, line 38: should we consider converting those parameters into option object instead? Later we can add other options and managing them via parameters might be painful

visual-python/src/saucelabs_visual/client.py
```diff
@@ -0,0 +1,185 @@
...
+                    deviceName
+                }
+            }
+            """
+        )
+        values = {"sessionId": session_id, "jobId": session_id}
+        meta = self.client.execute(query, variable_values=values)
+        return meta['meta']['blob']
+
+    def create_snapshot_from_webdriver(
```
konraddysput, line 126: 1. how about more generic function name? In the end it doesn't matter what is behind the method.
2. Should we consider passing some options via option object?

konraddysput (approved):

visual-python/src/saucelabs_visual/client.py
omacranger: This one specifically requires a running selenium session through Sauce for us to execute the backend snapshot tasks. More generic `create_snapshot` and `upload_snapshot` methods will likely come later once we broaden the framework support.

visual-python/src/saucelabs_visual/client.py
omacranger: They're all named parameters so they can be passed in an order or completely omitted. We also get decent typing like this without forcing users to instantiate `TypedDict`s just to pass an option. But I'm open to changing it if the whole team feels otherwise.

visual-python/src/saucelabs_visual/client.py
omacranger: Yep totally forgot to replace those after refactoring it. Thanks!

FriggaHel (approved):

PR #22, opened 2024-04-11
Build pipeline for Java SDK

kb-kerem:
Adds build pipeline for java sdk

### DIFF
--- a/.github/workflows/java-build.yml
+++ b/.github/workflows/java-build.yml
@@ -1,0 +1,28 @@
+name: Java (build)
+
+on:
+  push:
+    paths:
+      - 'visual-java/**'
+  pull_request:
+    paths:
+      - 'visual-java/**'
+
+defaults:
+  run:
+    working-directory: visual-java
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Java
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: '11'
+          cache: 'maven'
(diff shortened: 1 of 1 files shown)


.github/workflows/java-build.yml
```diff
@@ -0,0 +1,28 @@
+name: Java (build)
+
+on:
+  push:
+    paths:
+      - 'visual-java/**'
```
FriggaHel, line 6:
```suggestion
      - 'visual-java/**'
      - .github/workflows/java-build.yml
```
So the build will be kicked if we only modify the workflow (and won't discover that it fails upon merge)

.github/workflows/java-build.yml
```diff
@@ -0,0 +1,28 @@
...
+  pull_request:
+    paths:
+      - 'visual-java/**'
```
FriggaHel, line 9:
```suggestion
      - 'visual-java/**'
      - .github/workflows/java-build.yml
```

.github/workflows/java-build.yml
```diff
@@ -0,0 +1,28 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Java
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: '11'
```
FriggaHel, line 24: Shouldn't we compile it with Java8 since right now we expect to support starting version 8 ?

kb-kerem: Yes 8 makes more sense, will update

FriggaHel (approved):

PR #30, opened 2024-04-16
Python Updates / Additions

omacranger:
## Description
Adds a number of enhancements to the Python binding and Robot Framework integration.

- Adds ignore region selector & web element parsing for Selenium in Robot
- Adds method & keyword for getting build status and outputting the results in a CLI-compatible table
- Adds build logging for build creation and shutdown
- Sets up pytest and adds a handful of unit tests for new utils

## Types of Changes

- Refactor/improvements

### DIFF
--- a/visual-python/src/saucelabs_visual/regions.py
+++ b/visual-python/src/saucelabs_visual/regions.py
@@ -6,21 +6,27 @@ class Region:
     aliases: List[str] = []
     graphql_endpoint: str
     job_url_template: str
+    visual_url_template: str
 
     def __init__(
             self,
             name: str,
             aliases: List[str],
             graphql_endpoint: str,
-            job_url_template: str
+            job_url_template: str,
+            visual_url_template: str,
     ):
         self.name = name
         self.aliases = aliases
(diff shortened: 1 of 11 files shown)


visual-python/src/saucelabs_visual/regions.py
```diff
@@ -6,22 +6,28 @@ class Region:
...
         self.graphql_endpoint = graphql_endpoint
         self.job_url_template = job_url_template
+        self.visual_url_template = visual_url_template
 
     def job_url(self, job_id: str) -> str:
         return self.job_url_template.format(JOB_ID=job_id)
 
+    def build_url(self, build_id: str) -> str:
+        return self.visual_url_template.format(BUILD_ID=build_id)
+
```
FriggaHel, line 30: Shouldn't we rely on the `url` field that the API is returning ?

FriggaHel (approved):

visual-python/src/saucelabs_visual/regions.py
omacranger: Yep, makes sense to me. I'll update it

PR #32, opened 2024-04-19
Env var tests

kb-kerem:
# One-line summary

Activate env variable tests for java, dotnet and python

## Types of Changes


kb-kerem:
Will be broken down into separate MRs for each framework

PR #36, opened 2024-04-23
.NET: Add advanced diffing options

FriggaHel:
# Allow advanced diffing options

> Issue : IRIS-912

## Description

This PR adds support for Advanced Diffing / Ignore region options

## Types of Changes


visual-dotnet/SauceLabs.Visual/Models/SelectiveRegion.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace SauceLabs.Visual.Models
+{
+    public class SelectiveRegion
+    {
+        public IgnoreRegion? Region { get; set; }
+        public IWebElement? Element { get; set; }
+
+        public DiffingOption[]? EnableOnly { get; set; }
+        public DiffingOption[]? Disable { get; set; }
```
ebekebe, line 15: In JS we have called it `enableOnly` and `disableOnly`. Shall do the same here?

visual-dotnet/SauceLabs.Visual/Models/SelectiveRegion.cs
```diff
@@ -0,0 +1,32 @@
...
+            {
+                throw new VisualClientException("ambiguous selective region. Only one of 'region' or 'element' has to be specified.");
+            }
+            if (Region == null && Element == null)
+            {
+                throw new VisualClientException("invalid selective region. one of 'region' or 'element' has to be specified.");
+            }
+
+            var diffingOptions = DiffingOptionsInHelper.CreateFromEnableOnlyDisable(EnableOnly, Disable);
+            return Region != null ? new RegionIn(Region, diffingOptions) : new RegionIn(Element!, diffingOptions);
```
ebekebe, line 29: Maybe worth using something like this https://github.com/mcintyre321/OneOf. It sounds like that would make sure it is compile time checked.

visual-dotnet/SauceLabs.Visual/Utils/DiffingOptionsInHelper.cs
```diff
@@ -0,0 +1,68 @@
...
+        };
+
+        internal static DiffingOptionsIn? CreateFromEnableOnlyDisable(IReadOnlyCollection<DiffingOption>? enableOnly, IReadOnlyCollection<DiffingOption>? disable)
+        {
+            if (enableOnly?.Count > 0)
+            {
+                var options = new DiffingOptionsIn(false);
+                foreach (var opt in enableOnly)
+                {
+                    if (Mapping.TryGetValue(opt, out var fn))
```
ebekebe, line 45: Can't we just simply have an update method on `DiffingOptionsIn`? E.g. `options.set(opt, true)`. That should also remove the `Try...`.

visual-dotnet/SauceLabs.Visual/Models/SelectiveRegion.cs
FriggaHel: Of course we do, I was nearly sur I read `enableOnly` / `disable`, will address that.

visual-dotnet/SauceLabs.Visual/Utils/DiffingOptionsInHelper.cs
FriggaHel: I wanted to prevent the GraphQL layer to be aware of that enum that has been build on top since this is only providing customer a convenient way to select what to diff.

I can convert it to a `switch/case` may be a cleaner solution if you prefer this.

visual-dotnet/SauceLabs.Visual/Models/SelectiveRegion.cs
```diff
@@ -0,0 +1,41 @@
...
+            {
+                throw new VisualClientException("ambiguous selective region. Only one of 'region' or 'element' has to be specified.");
+            }
+            if (Region == null && Element == null)
+            {
+                throw new VisualClientException("invalid selective region. one of 'region' or 'element' has to be specified.");
+            }
+
+            var diffingOptions = DiffingOptionsInHelper.CreateFromEnableOnlyDisable(EnableOnly, DisableOnly);
+            return Region != null ? new RegionIn(Region, diffingOptions) : new RegionIn(Element!, diffingOptions);
```
ebekebe, line 38: I would rather make sure the compiler can check that `Element` is set:

```
if (Element != null && Region != null) throw ...
else if (Region != null) return new RegionIn(Region, diffingOptions);
else if (Element != null) return new RegionIn(Element, diffingOptions);
else throw ...
```

visual-dotnet/SauceLabs.Visual/Utils/DiffingOptionsInHelper.cs
ebekebe: I assume most users don't care about the underlying backend API. I think we should make the API that the user is actually uses safe and easy to use.

visual-dotnet/SauceLabs.Visual/Models/SelectiveRegion.cs
ebekebe: We can merge this now and change it later, as the API is still unstable /  beta.

ebekebe (approved):

PR #37, opened 2024-04-24
[fix] Env var precedence

kb-kerem:
# One-line summary

> Issue : IRIS-936

## Description
Fix build customizing environment variable handling for java and dotnet sdk.

## Types of Changes
- Breaking change (fix or feature that would cause existing functionality to change)

### DIFF
--- a/visual-java/src/main/java/com/saucelabs/visual/utils/EnvironmentVariables.java
+++ b/visual-java/src/main/java/com/saucelabs/visual/utils/EnvironmentVariables.java
@@ -1,5 +1,7 @@
 package com.saucelabs.visual.utils;
 
+import java.util.Optional;
+
 public class EnvironmentVariables {
 
   private EnvironmentVariables() {}
@@ -15,5 +17,14 @@ public class EnvironmentVariables {
 
   public static boolean isNotBlank(String str) {
     return str != null && !str.trim().isEmpty();
+  }
+
+  public static Optional<String> firstNonBlank(String... values) {
+    for (String value : values) {
+      if (isNotBlank(value)) {
+        return Optional.of(value);
+      }
+    }
+    return Optional.empty();
   }
 }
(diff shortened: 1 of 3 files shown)


visual-java/src/main/java/com/saucelabs/visual/utils/EnvironmentVariables.java
```diff
@@ -1,19 +1,31 @@
...
   private EnvironmentVariables() {}
 
-  public static final String PROJECT_NAME = System.getenv("SAUCE_VISUAL_PROJECT");
-  public static final String BRANCH_NAME = System.getenv("SAUCE_VISUAL_BRANCH");
-  public static final String DEFAULT_BRANCH_NAME = System.getenv("SAUCE_VISUAL_DEFAULT_BRANCH");
-  public static final String BUILD_NAME = System.getenv("SAUCE_VISUAL_BUILD_NAME");
-  public static final String BUILD_ID = System.getenv("SAUCE_VISUAL_BUILD_ID");
-  public static final String CUSTOM_ID = System.getenv("SAUCE_VISUAL_CUSTOM_ID");
+  public static final String PROJECT_NAME =
+      Optional.of(System.getenv("SAUCE_VISUAL_PROJECT")).orElse("");
```
FriggaHel, line 10: Is there a reason to default to empty string instead of null ?

kb-kerem: The idea is to not deal with nulls and also to provide some similarity to the way they're handled in dotnet.

FriggaHel: I'm not sure we should let languages specifications dict how implementation of another SDK should be done.

C# suggests to explicitly tell when a value can be nullable.
Since Java lets null to be passed, and is not an issue when serializing, it's probably better to stick with null as the empty value in that case. Wdyt ?

kb-kerem: There's no issue with serializing, that's right. Avoiding null here also provides a safe value to work with, see valueOrDefault would return a null when the second argument ends up being null. I'm not suggesting to use Java with C# standards but this specific inspiration does bring value IMHO.

FriggaHel (approved):

PR #38, opened 2024-04-24
IRIS-927-align-java-fps-options

paweltomaszewskisaucelabs:
## Description
Add new fps options into java bindings

## Types of Changes
Configuration change

### DIFF
--- a/visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
+++ b/visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
@@ -199,9 +199,7 @@ public class VisualApi {
     this.captureDom = captureDom;
   }
 
-  /**
-   * Enables full page screenshots
-   */
+  /** Enables full page screenshots */
   public void enableFullPageScreenshots() {
     this.fullPageScreenshotConfig = new FullPageScreenshotConfig.Builder().build();
   }
@@ -385,7 +383,9 @@ public class VisualApi {
       input.setClipSelector(clipSelector);
     }
 
-    FullPageScreenshotConfig fullPageScreenshotConfig = Optional.ofNullable(options.getFullPageScreenshotConfig()).orElse(this.fullPageScreenshotConfig);
+    FullPageScreenshotConfig fullPageScreenshotConfig =
+        Optional.ofNullable(options.getFullPageScreenshotConfig())
+            .orElse(this.fullPageScreenshotConfig);
     if (fullPageScreenshotConfig != null) {
       input.setFullPageConfig(fullPageScreenshotConfig);
     }
--- a/visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
+++ b/visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
@@ -95,7 +95,7 @@ public class VisualApi {
         api.setCaptureDom(this.captureDom);
       }
       if (this.fullPageScreenshotConfig != null) {
-        api.setFullPageScreenshotConfig(this.fullPageScreenshotConfig);
+        api.enableFullPageScreenshots(this.fullPageScreenshotConfig);
       }
       return api;
     }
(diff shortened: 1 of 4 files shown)


visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -190,6 +201,15 @@ public void setCaptureDom(Boolean captureDom) {
     this.captureDom = captureDom;
   }
 
+  /**
+   * Enables full page screenshots
+   *
+   * @param fullPageScreenshotConfig activation of DOM Capture.
```
FriggaHel, line 207: Not sure to understand what that means.

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -370,7 +390,10 @@ public void sauceVisualCheck(String snapshotName, CheckOptions options) {
       input.setClipSelector(clipSelector);
     }
 
-    input.setFullPageConfig(options.getFullPageScreenshotConfig());
+    FullPageScreenshotConfig fullPageScreenshotConfig = Optional.ofNullable(options.getFullPageScreenshotConfig()).orElse(this.fullPageScreenshotConfig);
+    if (fullPageScreenshotConfig != null) {
+      input.setFullPageConfig(fullPageScreenshotConfig);
+    }
```
FriggaHel, line 396: Wouldn't something like this be simplier ?
```suggestion
    input.setFullPageConfig(
      Optional.ofNullable(options.getFullPageScreenshotConfig()).orElse(this.fullPageScreenshotConfig)
    );
```

kb-kerem: I think the intention is not to set it on `input` if `this.fullPageScreenshotConfig` is null. It doesn't read great, but it follows the way it was done `captureDom`

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -190,6 +201,15 @@ public void setCaptureDom(Boolean captureDom) {
     this.captureDom = captureDom;
   }
 
+  /**
+   * Enables full page screenshots
+   *
+   * @param fullPageScreenshotConfig activation of DOM Capture.
+   */
+  public void setFullPageScreenshotConfig(FullPageScreenshotConfig fullPageScreenshotConfig) {
```
kb-kerem, line 209: I see that this is aligned with the way captureDom is handled but in CheckOptions there are two functions provided for enabling FPS called `enableFullPageScreenshots()` and `enableFullPageScreenshots(FullPageScreenshotConfig config)`. This arguably reads nicer (as it's a user facing API) and provides a way to enable basic FPS without passing a config. I suggest we do the same here

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
paweltomaszewskisaucelabs: good catch!

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
paweltomaszewskisaucelabs: Changed. Good idea, thanks.

visual-java/src/main/java/com/saucelabs/visual/model/FullPageScreenshotConfig.java
```diff
@@ -6,29 +6,68 @@
...
 
+    public Builder withHideScrollBars(Boolean hideScrollBars) {
+      this.hideScrollBars = hideScrollBars;
+      return this;
+    }
+
+    public Builder withScrollLimit(int scrollLimit) {
+      if (scrollLimit > 0 && scrollLimit <= 10) {
+        this.scrollLimit = scrollLimit;
+      } else {
```
FriggaHel, line 57: This should be enforced only by the backend.

If for some reason we raise the scrollLimit, we will have to update the client which is unnecessary burden.

Client should only check that this is a positive value.

FriggaHel (approved):

PR #42, opened 2024-05-02
Java: Add advanced diffing options

FriggaHel:
This PR adds support for Advanced Diffing / Ignore region options

## Types of Changes


visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -371,12 +375,53 @@ public void sauceVisualCheck(String snapshotName, CheckOptions options) {
...
+    return builder;
+  }
+
+  private DiffingOptionsIn generateDiffingOptions(
+      List<String> enableOnly, List<String> disableOnly) {
+    if (enableOnly != null && !enableOnly.isEmpty()) {
+      DiffingOptionsIn.Builder builder = DiffingOptionsIn.builder();
+      for (String option : DiffingOptionValues) {
+        setDiffingOptionValue(builder, option, enableOnly.contains(option));
+      }
```
ebekebe, line 403: Why not just loop through `enableOnly`?

Also, if you check if both, `enableOnly` and `disableOnly` are null, you can return early. That way the builder can be shared between disableOnly and enableOnly. Saves 2 lines.

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -371,12 +375,53 @@ public void sauceVisualCheck(String snapshotName, CheckOptions options) {
...
+        return builder.withStyle(value);
+      case Visual:
+        return builder.withVisual(value);
+    }
+    return builder;
+  }
+
+  private DiffingOptionsIn generateDiffingOptions(
+      List<String> enableOnly, List<String> disableOnly) {
+    if (enableOnly != null && !enableOnly.isEmpty()) {
```
ebekebe, line 399: if `enableOnly` is empty, it should create DiffingOptions with all flags set to `false` (same for `disableOnly`).

ebekebe (changes requested):

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
FriggaHel: If I just loop with `enableOnly`, the extraneous values will be `null` not `false`.

ebekebe (approved):

PR #53, opened 2024-05-20
Implement selective diffing in Cypress

FriggaHel:
# One-line summary

> Issue : IRIS-815

## Description

Add supports for Selective Diffing in Cypress

## Types of Changes


kb-kerem:
changes are ok but we need a changeset file which needs to be generated by running `npx changeset` and selecting the packages to be bumped

kb-kerem (changes requested):
changes are ok but we need a changeset file which needs to be generated by running `npx changeset` and selecting the packages to be bumped

PR #57, opened 2024-05-21
fps deprecate some options

paweltomaszewskisaucelabs:
## Description
deprecate some fps options

### DIFF
--- a/visual-js/.changeset/loud-timers-check.md
+++ b/visual-js/.changeset/loud-timers-check.md
@@ -1,0 +1,7 @@
+---
+"@saucelabs/nightwatch-sauce-visual-service": minor
+"@saucelabs/wdio-sauce-visual-service": minor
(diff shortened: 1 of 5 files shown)


visual-js/.changeset/proud-wolves-argue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@saucelabs/visual": patch
```
FriggaHel, line 2: Shouldn't deprecation be a minor update rather than a patch ?

FriggaHel: According to semver, it should: https://semver.org/#spec-item-7

paweltomaszewskisaucelabs: ok, I changed it

FriggaHel (approved):

PR #63, opened 2024-05-23
[JS/WDIO] WebElement API support

omacranger:
## Description
- Adds support for passing WebElements & WebElementArrays into the WDIO bindings.
- Updates / fixes the GraphQL codegen for open source JS SDKs
- Fixes / updates the exported types for Nightwatch.

## Types of Changes

- Refactor/improvements


visual-js/visual-wdio/src/SauceVisualService.ts
```diff
@@ -163,86 +147,65 @@ ${url.padStart(100, ' ')}
...
-          `Ignored element options.ignore[${i}] is not valid: ${e}`,
-        );
-      }
-    }
+const getDiffingOptions = (
+  region: RegionType,
+): DiffingOptionsIn | undefined => {
+  const { element: __, ...rest } = region;
+  if ('enableOnly' in rest || 'disableOnly' in rest) {
+    return selectiveRegionOptionsToDiffingOptions(rest);
```
ebekebe, line 155: Is there specific reason why you want to let the backend handle the default instead of `selectiveRegionOptionsToDiffingOptions`?

If the behaviour is not satisfactory, we should change `selectiveRegionOptionsToDiffingOptions` directly and not wrap it.

visual-js/visual-wdio/src/SauceVisualService.ts
```diff
@@ -163,86 +147,65 @@ ${url.padStart(100, ' ')}
...
+  const ignoreElements: ElementIn[] = [];
+  const ignoreRegions: RegionIn[] = [];
+
+  const pushElement = async (region: RegionType) => {
+    const { element: maybePromiseElement, ...other } = region;
+    const element = await maybePromiseElement;
+
+    if (Array.isArray(element)) {
+      for await (const subElement of element) {
+        await pushElement({
```
ebekebe, line 181: I am not sure if the recursion is necessary: Are we dealing with a deep tree here? I thought it is depth 2, i.e. maps to `RegionIn[][]`.

I would also try an linearize the code, by converting to an array if necessary:

```ts
const ignoreRegions = Promise.all(ignoreElements.maps(async region => {
  const {element, ...other } = await region.element;
  const subElements = Array.isArray(element) ? element : [element];
  return subElements.map(sub => ({
    ...(isWdio(sub) ? {id: ....} : {...sub}),
    ...other,
    diffingOptions: getDiffingOptions(region),
  }))
})).flat()
```

Since this code is similar to in all JS frameworks, we might want to move it the `visual` package and only keep the detection and conversion of wdio elements in here. (Not sure about this one.)

visual-js/visual/src/utils.ts
```diff
@@ -1,4 +1,5 @@
-import { FullPageConfigIn } from './graphql/__generated__/graphql';
+import { type, Type } from 'arktype';
```
ebekebe, line 1: We used `zod` in the past. What is the benefit of `arktype`?

visual-js/visual-wdio/src/SauceVisualService.ts
omacranger: It's less for us to maintain long term if we let the backend handle as much of the defaults as we can.

visual-js/visual-wdio/src/SauceVisualService.ts
omacranger: We'd just have to then re-filter those back out into separate arrays. 

The elements used by Nightwatch also vary. WDIO uses, well, WDIO, where Nightwatch AFAIK uses selenium underneath which has different element types. Though I suppose that could be exported and allowing `isWdio` (is webelement, maybe) to be customized.

visual-js/visual/src/utils.ts
omacranger: You tell me, it was originally your addition to `visual-wdio` I'm just moving it into the Visual package 😅

ebekebe: forgot about that 🤦‍♂️

visual-js/visual/src/utils.ts
```diff
@@ -20,3 +28,92 @@ export const isSkipMode = (): boolean => {
...
+      });
+    } else if (isIgnoreRegion(element)) {
+      ignoreRegions.push({
+        ...element,
+        diffingOptions: getDiffingOptions(region),
+      });
+    }
+  };
+
+  for await (const region of ignore) {
```
ebekebe, line 111: potentially unsafe as `ignore` is an array: https://stackoverflow.com/questions/59694309/for-await-of-vs-promise-all

```suggestion
  for (const region of Promise.all(ignore)) {
```

visual-js/visual/src/utils.ts
```diff
@@ -20,3 +28,92 @@ export const isSkipMode = (): boolean => {
...
+    const { element: maybePromiseElement, ...other } = region;
+    const element = await maybePromiseElement;
+
+    if (Array.isArray(element)) {
+      for await (const subElement of element) {
+        await pushElement({
+          element: subElement,
+          ...other,
+        });
+      }
```
ebekebe, line 93: There is still the recursion that can be eliminated (to make it easier for future editors and reviewers), because we only have two nested layers.

I also think, we only need to pass one function that checks and resolves at the same time:

```ts
resolveElement: (maybeElement: unknown) => Promise<{id: string, name?: string}[] | null>
```

visual-js/visual-nightwatch/src/nightwatch/commands/sauceVisualCheck.ts
```diff
@@ -65,12 +74,49 @@ export default class SauceVisualCheck implements NightwatchCustomCommandsModel {
...
+          const awaited = await element;
+          return {
+            ...other,
+            // If the items are an array, await all of them to get the resolved WebElement promises instead.
+            element: Array.isArray(awaited)
+              ? await Promise.all(awaited)
+              : awaited,
+          } satisfies RegionType<ParsedElementType>;
+        }),
+    );
```
ebekebe, line 111: Why not throw everything into `parseRegionsForAPI` without so much pre-processing and extract the elements in `getElementMeta`? It needs to return an array (see comment below) and that this code should be much simpler.

omacranger: _Some_ preprocessing is going to be required since we have some tooling that currently supports selectors, like nightwatch, but isn't supported in WDIO. But sure we can simplify those and potentially do more insid parseRegionsForAPI.

visual-js/visual/src/utils.ts
```diff
@@ -96,12 +99,15 @@ export const parseRegionsForAPI = async <T>(
...
-      const { item, diffingOptions } = isRegionType(itemOrRegion)
-        ? {
-            item: itemOrRegion.element,
-            diffingOptions: getDiffingOptions(itemOrRegion),
-          }
-        : { item: itemOrRegion, diffingOptions: undefined };
+      const { item, diffingOptions } =
+        // Can only use `in` operators (arktype validation) on objects. Strings passed here
+        // would error.
+        typeof itemOrRegion === 'object' && isRegionType(itemOrRegion)
```
ebekebe, line 105: Can the `object` check be moved into `isRegionType`?

ebekebe (approved):

PR #64, opened 2024-05-24
[Java] Add backend WebElement support

omacranger:
## Description
Adds WebElement support to the java bindings to reduce frontend calculation.

## Types of Changes

- Refactor/improvements

### DIFF
--- a/visual-java/src/main/graphql/visual/schema.graphqls
+++ b/visual-java/src/main/graphql/visual/schema.graphqls
@@ -1,20 +1,20 @@
 schema {
-  mutation: Mutation
   query: Query
+  mutation: Mutation
+}
+
+"An object with a globally unique `ID`."
+interface Node {
+  "A globally unique identifier. Can be used in various places throughout the system to identify this single value."
+  nodeId: ID!
 }
 
 type ApplicationSummary {
   name: String!
 }
 
-input ApproveBuildIn {
-  """@deprecated Use `uuid`. This field will be removed in a future update."""
-  id: ID
-  onlyNew: Boolean
-  uuid: UUID
-}
-
 """
+
 A `Baseline` is what a `Snapshot` is compared to.
 
 For details see https://docs.saucelabs.com/visual-testing/sauce-visual/index.html#baseline-matching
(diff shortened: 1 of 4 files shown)


visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -520,13 +512,39 @@ private List<RegionIn> extractIgnoreList(CheckOptions options) {
     return result;
   }
 
+  private List<ElementIn> extractIgnoreElements(CheckOptions options) {
+    if (options == null) {
+      return Collections.emptyList();
+    }
+
+    List<WebElement> ignoredElements =
+            options.getIgnoreElements() == null ? Arrays.asList() : options.getIgnoreElements();
```
ebekebe, line 521: [minor]
```suggestion
    List<WebElement> ignoredElements = options != null && options.getIgnoreElements() != null
            ? options.getIgnoreElements()
            : Arrays.asList();
```

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -520,13 +512,39 @@ private List<RegionIn> extractIgnoreList(CheckOptions options) {
...
     }
     return element;
   }
 
+  private WebElement validateRemoteElement(WebElement element) {
+    if (element == null || !element.isDisplayed() || !(element instanceof RemoteWebElement)) {
+      return null;
+    }
+    return element;
+  }
```
ebekebe, line 546: maybe easier to read
```suggestion
    if (element != null && element instanceof RemoteWebElement && element.isDisplayed()) {
      return element;
    }
    return null;
  }
```

ebekebe (approved):

visual-java/src/main/graphql/visual/schema.graphqls
```diff
@@ -1,20 +1,20 @@
 schema {
-  mutation: Mutation
```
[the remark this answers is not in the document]
omacranger: Ah, yes I didn't copy from the backend schema. I dumped it using a GraphQL client. I'll copy it to make it more consistent -- but perhaps we should put a CLI or script here to dump that instead like the JS clients?

PR #66, opened 2024-05-27
Add support for SAUCE_VISUAL_SKIP in Nightwatch

FriggaHel:
# Add support for SAUCE_VISUAL_SKIP in Nightwatch

> Issue : IRIS-1000

## Description

Add support for `SAUCE_VISUAL_SKIP` in Nightwatch

## Types of Changes


konraddysput (approved):

visual-js/visual-nightwatch/nightwatch/assertions/sauceVisualResults.js
```diff
@@ -58,6 +58,17 @@ exports.assertion = function sauceVisualResults(diffStatus, expected, msg) {
...
   this.command = async function (callback) {
+    // Return only SKIPPED if in skip mode
+    if (isSkipMode()) {
+      const statusSummary = Object.fromEntries(
+        Object.values(DiffStatus).map((status) => [status, 0]),
+      );
+      statusSummary['SKIPPED'] = global.skipped;
+      return callback({
+        value: statusSummary[diffStatus],
+      });
```
ebekebe, line 69: I don't get this code. In the end only `statusSummary[diffStatus]` is used. According to my understanding of the code this value is always `0` unless, `diffStatus == 'SKIPPED'`.

What is it supposed to do?

visual-js/visual/src/utils.ts
```diff
@@ -15,3 +15,8 @@ export const getFullPageConfig: (
   const localCfg = typeof local === 'object' ? local : {};
   return { ...globalCfg, ...localCfg };
 };
+
+export const isSkipMode = (): boolean => {
+  const VISUAL_SKIP_ENV_VAR = 'SAUCE_VISUAL_SKIP';
+  return !!process.env[VISUAL_SKIP_ENV_VAR];
```
ebekebe, line 21:
```suggestion
  return Boolean(process.env[VISUAL_SKIP_ENV_VAR]);
```

more idiomatic (these days)

visual-js/visual-nightwatch/nightwatch/assertions/sauceVisualResults.js
omacranger: Looks like it's supposed to emulate a `getVisualResults` check so they can assert against it (ex: assert that we have 1 unapproved snapshot after check - `sauceVisualResults(DiffStatus.Unapproved, 1)`). IMO this command should probably be deprecated / removed since it's bad practice while the diffs are async. Though, to keep the functionality as-is while adding skipped behavior it should probably be below to match the nightwatch builder API

```
callback({
    value: statusSummary[diffStatus],
});
return this;
```

visual-js/.changeset/wet-pianos-work.md
```diff
@@ -0,0 +1,6 @@
+---
+"@saucelabs/visual": minor
+"@saucelabs/nightwatch-sauce-visual-service": minor
```
omacranger, line 3: Are we sure these should be 'minor'? In JS ecosystem `minor` bumps in a prerelease version (`0.x.x`) would not be considered 'compatible'

ebekebe (approved):

PR #83, opened 2024-06-26
[java] Support SAUCE_REGION

kb-kerem:
> Issue : IRIS-1069

## Description
Bring SAUCE_REGION support to Java SDK

## Types of Changes
- Refactor/improvements

### DIFF
--- a/visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
+++ b/visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
@@ -236,13 +236,15 @@ public class VisualApi {
     WebdriverSessionInfoQuery query =
         new WebdriverSessionInfoQuery(
             new WebdriverSessionInfoQuery.WebdriverSessionInfoIn(this.jobId, this.sessionId));
-    WebdriverSessionInfoQuery.Data response =
-        this.client.execute(query, WebdriverSessionInfoQuery.Data.class);
-    if (response.result == null) {
-      throw new VisualApiException(
+    try {
+      WebdriverSessionInfoQuery.Data response =
+          this.client.execute(query, WebdriverSessionInfoQuery.Data.class);
+      return response.result;
+    } catch (VisualApiException e) {
+      log.error(
           "Sauce Visual: No WebDriver session found. Please make sure WebDriver and Sauce Visual data centers are aligned.");
+      throw e;
     }
-    return response.result;
   }
 
   /**
--- a/visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
+++ b/visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
@@ -25,7 +25,7 @@ import org.slf4j.LoggerFactory;
 public class VisualApi {
   private static final Logger log = LoggerFactory.getLogger(VisualApi.class);
 
-  public static String resolveEndpoint(String endpoint) {
+  private static String resolveEndpoint(String endpoint) {
     if (isNotBlank(endpoint)) {
       return endpoint;
     }
--- a/visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
+++ b/visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
@@ -25,6 +25,13 @@ import org.slf4j.LoggerFactory;
 public class VisualApi {
   private static final Logger log = LoggerFactory.getLogger(VisualApi.class);
 
+  public static String resolveEndpoint(String endpoint) {
+    if (isNotBlank(endpoint)) {
+      return endpoint;
+    }
+    return DataCenter.fromSauceRegion(EnvironmentVariables.SAUCE_REGION).endpoint;
+  }
+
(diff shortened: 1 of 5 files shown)


visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -25,6 +25,13 @@
 public class VisualApi {
   private static final Logger log = LoggerFactory.getLogger(VisualApi.class);
 
+  public static String resolveEndpoint(String endpoint) {
```
konraddysput, line 28: do we need to expose this option here? It doesn't look like a correct domain for VisualApi

kb-kerem: right, this can be certainly made private

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -50,7 +57,7 @@ public Builder(RemoteWebDriver driver, String username, String accessKey, String
       this.driver = driver;
       this.username = username;
       this.accessKey = accessKey;
-      this.endpoint = endpoint;
+      this.endpoint = resolveEndpoint(endpoint);
```
FriggaHel, line 60: `resolveEndpoint` should probably take a look into `driver` to see if it can infer the actual region in use.

It will prevent user setups where session runs on region `A` but Visual build is on region `B`.

kb-kerem: Yeah that would be ideal and simplify our constructor and usage greatly. Unfortunately there's no way to infer the actual region from `driver` at the moment.
See: https://www.selenium.dev/selenium/docs/api/java/org/openqa/selenium/remote/RemoteWebDriver.html

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -112,6 +119,17 @@ public VisualApi build() {
...
 
+  /**
+   * Creates a VisualApi instance for a given Visual Backend {@link DataCenter}
+   *
+   * @param driver The {@link org.openqa.selenium.WebDriver} instance where the tests should run at
+   * @param username SauceLabs username
+   * @param accessKey SauceLabs access key
+   */
+  public VisualApi(RemoteWebDriver driver, String username, String accessKey) {
+    this(driver, "", username, accessKey);
```
FriggaHel, line 130: Wouldn't it be better if this is resolved prior invoking the other constructor ?

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -160,7 +178,7 @@ public VisualApi(
           "Invalid SauceLabs credentials. "
               + "Please check your SauceLabs username and access key at https://app.saucelabs.com/user-settings");
     }
-    this.client = new GraphQLClient(url, username, accessKey);
+    this.client = new GraphQLClient(resolveEndpoint(url), username, accessKey);
```
FriggaHel, line 181: As mentioned in my other comment, I'm not super convinced by invoking `resolveEndpoint` here.

IMO `resolveEndpoint` should only be invoked when there is not value available (I mean when the function invoked does not have this param).

Any opinion @konraddysput ?

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -218,7 +236,13 @@ private WebdriverSessionInfoQuery.Result webdriverSessionInfo() {
     WebdriverSessionInfoQuery query =
         new WebdriverSessionInfoQuery(
             new WebdriverSessionInfoQuery.WebdriverSessionInfoIn(this.jobId, this.sessionId));
-    return this.client.execute(query, WebdriverSessionInfoQuery.Data.class).result;
+    WebdriverSessionInfoQuery.Data response =
+        this.client.execute(query, WebdriverSessionInfoQuery.Data.class);
+    if (response.result == null) {
+      throw new VisualApiException(
```
FriggaHel, line 242: Can we also log the `traceId` there ?

It will be helpful to troubleshoot customers

FriggaHel (changes requested):

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
kb-kerem: I agree that this doesn't read great, however this implementation is due to a limitation of java regarding usage of `this` keyword. It has be the first thing executed when used in a constructor and no other function calls are allowed inside the statement. To modify this behaviour we need to rework the constructor calling logic, which is probably not something we want to do for this task

kb-kerem: Turns out that static functions are indeed callable using `this` in constructor. 🤦 
It should look better now.

FriggaHel (approved):

PR #89, opened 2024-07-02
Fix js SDK client version setting in release pipeline

kb-kerem:
> Issue : IRIS-1076

## Description
Fixes the broken behavior of setting PKG_VERSION in js SDKs 

## Types of Changes

### DIFF
--- a/visual-js/replace_pkg_version.sh
+++ b/visual-js/replace_pkg_version.sh
@@ -4,7 +4,7 @@
 if ! command -v jq &> /dev/null
 then
     echo "jq could not be found. Please install jq to run this script."
-    exit
+    exit 1
 fi
 
 # Check if a JSON file is provided as an argument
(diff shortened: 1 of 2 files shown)


visual-js/replace_pkg_version.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+# Check if jq is installed
+if ! command -v jq &> /dev/null
+then
+    echo "jq could not be found. Please install jq to run this script."
+    exit
```
FriggaHel, line 7:
```suggestion
    exit 1
```

Otherwise failures may not be catched.

visual-js/replace_pkg_version.sh
```diff
@@ -0,0 +1,68 @@
...
+JSON_FILE=$1
+
+# Check if the JSON file exists
+if [ ! -f "$JSON_FILE" ]; then
+    echo "JSON file not found!"
+    exit 1
+fi
+
+# Define the mapping between release names and filenames
+release_names=(
```
FriggaHel, line 25: Why not using associative arrays ? https://www.gnu.org/software/bash/manual/html_node/Arrays.html

It may be slightly less error-prone.

Note: Bash5 is ideal, bash4 is the minimum version needed.

FriggaHel (changes requested):

visual-js/replace_pkg_version.sh
kb-kerem: good point. done

visual-js/replace_pkg_version.sh
kb-kerem: I wasn't sure if bash4 is on the runner. Updated.

FriggaHel: You can go to the safe side by adding this at the top of the file:

```
if [ "${BASH_VERSINFO[0]}" -lt 5 ];then
  echo "Bash5 is required" > /dev/stderr
  exit 1
fi
```

FriggaHel (approved):
LGTM.

FYI: [Bash5 is installed by default in `ubuntu-latest` image](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md)

PR #101, opened 2024-08-01
[java] clip element support

kb-kerem:
> Issue : IRIS-1140

## Description
Adds clip to element support to java.

## Types of Changes

### DIFF
--- a/visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
+++ b/visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
@@ -16,6 +16,7 @@ import dev.failsafe.RetryPolicy;
 import java.time.Duration;
 import java.util.*;
 import java.util.stream.Collectors;
+import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.remote.RemoteWebDriver;
 import org.openqa.selenium.remote.RemoteWebElement;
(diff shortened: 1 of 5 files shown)


FriggaHel:
> NOTE: Conversion of clipSelector to clipElement is not included since wd is not available at createSnapshot time, unlike wdio

Shouldn't we simply store the driver instance that we get in the builder to address this ?

kb-kerem:
> > NOTE: Conversion of clipSelector to clipElement is not included since wd is not available at createSnapshot time, unlike wdio
> 
> Shouldn't we simply store the driver instance that we get in the builder to address this ?

Yep actually that makes sense. Added it.

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -176,7 +177,8 @@ public VisualApi(
               + "Please check your SauceLabs username and access key at https://app.saucelabs.com/user-settings");
     }
     this.client = new GraphQLClient(url, username, accessKey);
-    this.sessionId = driver.getSessionId().toString();
```
FriggaHel, line 179: Shouldn't we keep this one too. We always require a string, while driver provides a `SessionId`.

IMO It will be more readable to have `this.sessionId` rather than `this.driver.getSessionId().toString()`

WDYT?

kb-kerem: Yeah I was not sure about removing it either. Added it.

FriggaHel (approved):

PR #102, opened 2024-08-02
[nightwatch] clip element support

kb-kerem:
> Issue : IRIS-1143

## Description
clipSelector support for Nightwatch

## Types of Changes

### DIFF
--- a/visual-js/visual-nightwatch/integration-tests/nightwatch/tests/saucedemo.ts
+++ b/visual-js/visual-nightwatch/integration-tests/nightwatch/tests/saucedemo.ts
@@ -2,7 +2,9 @@ import { NightwatchTests } from 'nightwatch';
 
 const home: NightwatchTests = {
   'Check Login Page': () => {
-    browser.url('https://saucedemo.com').sauceVisualCheck('Login page');
+    browser.url('https://saucedemo.com').sauceVisualCheck('Login page', {
+      clipElement: browser.element('input[data-test="login-button"]'),
+    });
   },
 };
 
--- a/visual-js/visual-nightwatch/integration-tests/nightwatch/tests/saucedemo.ts
+++ b/visual-js/visual-nightwatch/integration-tests/nightwatch/tests/saucedemo.ts
@@ -2,6 +2,9 @@ import { NightwatchTests } from 'nightwatch';
 
 const home: NightwatchTests = {
   'Check Login Page': () => {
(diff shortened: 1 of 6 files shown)


visual-js/visual-nightwatch/src/nightwatch/commands/sauceVisualCheck.ts
```diff
@@ -132,6 +132,10 @@ class SauceVisualCheck implements NightwatchCustomCommandsModel {
       ReturnType<typeof api.createSnapshotFromWebDriver>
     > | null = null;
 
+    const clipElement = options.clipSelector
```
FriggaHel, line 135: I would probably rename this to that:
```suggestion
    const clipElementFromClipSelector = options.clipSelector
```

It will help gain clarity on line 151:

From
```
clipElement: (await options.clipElement?.getId()) ?? clipElement,
```

to
```
clipElement: (await options.clipElement?.getId()) ?? clipElementFromClipSelector,
```

Otherwise you are in a situation where `option.clipElement` and `clipElement` are not as similar as naming could suggest.

FriggaHel:
Additionnally, can we add it to the integration tests please ?

visual-js/visual-nightwatch/integration-tests/nightwatch/tests/saucedemo.ts
```diff
@@ -2,7 +2,9 @@ import { NightwatchTests } from 'nightwatch';
 
 const home: NightwatchTests = {
   'Check Login Page': () => {
-    browser.url('https://saucedemo.com').sauceVisualCheck('Login page');
+    browser.url('https://saucedemo.com').sauceVisualCheck('Login page', {
```
omacranger, line 5: This should probably be in a separate test so we can keep the default functionality as is?

omacranger (approved):
Just the one comment, otherwise looks good to me

visual-js/visual-nightwatch/integration-tests/nightwatch/tests/saucedemo.ts
FriggaHel: I Agree.

Ideally we should be able to accept baseline for both situations and ensure this is always equal.

kb-kerem: moved it to a separate test

PR #104, opened 2024-08-15
[C#] Add baselineOverride functionality

omacranger:
## Description
Adds the ability to override the baseline matching behavior at a build or visual check level.

## Types of Changes

### DIFF
--- a/visual-dotnet/SauceLabs.Visual/BaselineOverride.cs
+++ b/visual-dotnet/SauceLabs.Visual/BaselineOverride.cs
@@ -3,6 +3,11 @@ using SauceLabs.Visual.Models;
 
 namespace SauceLabs.Visual
 {
+    /// <summary>
+    /// <c>BaselineOverride</c> One or more values from 'SnapshotIn' we should use as an override
+    /// when finding a baseline. Enables cross browser / OS snapshot comparison by matching the
+    /// values set on the device you'd like to compare against.
+    /// </summary>
     public class BaselineOverride
     {
         public Browser? Browser { get; set; }
--- a/visual-dotnet/SauceLabs.Visual/BaselineOverride.cs
+++ b/visual-dotnet/SauceLabs.Visual/BaselineOverride.cs
@@ -1,0 +1,32 @@
+using SauceLabs.Visual.GraphQL;
+using SauceLabs.Visual.Models;
+
+namespace SauceLabs.Visual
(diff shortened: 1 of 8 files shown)


visual-dotnet/SauceLabs.Visual/BaselineOverride.cs
```diff
@@ -0,0 +1,32 @@
+using SauceLabs.Visual.GraphQL;
+using SauceLabs.Visual.Models;
+
+namespace SauceLabs.Visual
+{
+    public class BaselineOverride
```
konraddysput, line 6: Can we add a summary what it is? Since this is a public interface, we need to provide some explanation what it does

visual-dotnet/SauceLabs.Visual/Models/Browser.cs
```diff
@@ -0,0 +1,11 @@
+namespace SauceLabs.Visual.Models
+{
+    public enum Browser
+    {
+        Chrome,
```
konraddysput, line 5: enum will be translated into a number. Are you sure the same order is on the server side? Maybe we can set a value to each option to be sure it will always be converted into a correct value (like in the future some adds lets say Internet explorer in the middle of the enum value)

visual-dotnet/SauceLabs.Visual/Models/OperatingSystem.cs
```diff
@@ -0,0 +1,11 @@
+namespace SauceLabs.Visual.Models
+{
+    public enum OperatingSystem
+    {
+        Android,
```
konraddysput, line 5: like above - should we set a value for each option to be sure the backend will interpret it always the same way?

konraddysput (approved):

visual-dotnet/SauceLabs.Visual/Models/Browser.cs
omacranger: Our GraphQL tooling converts these to CONSTANT_CASE to match our API, see [here](https://graphql-dotnet.github.io/docs/getting-started/schema-types/#enumerations).

visual-dotnet/SauceLabs.Visual/BaselineOverride.cs
omacranger: Done!

PR #107, opened 2024-08-20
[Python] Baseline Overrides / Mobile support

omacranger:
## Description
Adds a few updates to the Visual Python & Robot Framework binding:

- Adds support for baseline overrides / cross browser / os comparisons
- Adds the ability to configure global client options to avoid setting the same setting on every snapshot for the following
  - dom capture
  - baseline overrides
  - diffing method
  - full page config
- Adds RDC support for SauceLabs by forwarding the Job ID from the driver results

## Types of Changes

- Refactor/improvements

## Deployment Notes
These should highlight any db migrations, feature toggles, etc.

### DIFF
--- a/visual-python/Dockerfile
+++ b/visual-python/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.12-slim-bookworm
+FROM python:3.11-slim-bookworm
 
 COPY . .
 
--- a/visual-python/src/saucelabs_visual/frameworks/robot.py
+++ b/visual-python/src/saucelabs_visual/frameworks/robot.py
@@ -7,6 +7,7 @@ from robot.api import logger
 from robot.api.deco import library, keyword
 from robot.libraries.BuiltIn import BuiltIn
 from selenium.webdriver.remote.webelement import WebElement
+from selenium.webdriver.remote.webdriver import WebDriver as RemoteWebDriver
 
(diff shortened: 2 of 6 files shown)


visual-python/Dockerfile
```diff
@@ -1,9 +1,9 @@
-FROM python:3.12-slim-bookworm
+FROM python:3.11-slim-bookworm
```
omacranger, line 1: Reduce from 3.12 -> 3.11 since 3.12 is not officially supported by robotframework-seleniumlibrary: https://github.com/robotframework/SeleniumLibrary/blob/master/setup.py#L44

visual-python/src/saucelabs_visual/frameworks/robot.py
```diff
@@ -265,7 +304,7 @@ def visual_snapshot(
 
         self.client.create_snapshot_from_webdriver(
             name=name,
-            session_id=session_id,
+            driver=self._get_driver(),
```
omacranger, line 307: This is technically a breaking change if anyone is using the current framework-free Python implementation. We're requiring the job ID here anyway -- so it would require manual intervention regardless. I figured it would be better to do this now and bypass all of that and just do a minor bump since this is prerelease.

kb-kerem (approved):

PR #115, opened 2024-08-28
Add command to merge baselines

ebekebe:
> Issue : IRIS-1166

## Description

This allows to copy baselines from a source branch to a target branch. This is useful to copy baselines from a feature branch to the main branch once the feature branch is merged.

### DIFF
--- a/visual-js/visual/src/cli/baselines.ts
+++ b/visual-js/visual/src/cli/baselines.ts
@@ -1,0 +1,56 @@
+import { Command, CommandOptions } from 'commander';
+import { VisualApiRegion } from '../regions';
+import { VisualApi, getApi } from '../api';
+
+function apiFromOptions(opts: { region: VisualApiRegion }): VisualApi {
+  return getApi({
+    region: opts.region,
+    user: process.env.SAUCE_USERNAME,
+    key: process.env.SAUCE_ACCESS_KEY,
+  });
+}
+
+type BuildStatusCommandOptions = {
+  project: string;
+  sourceBranch: string;
+  targetBranch: string;
+  region: VisualApiRegion;
+};
+export async function baselineMergeCommand(_: unknown, cmd: Command) {
+  const options = cmd.optsWithGlobals<BuildStatusCommandOptions>();
+
+  const visualApi = apiFromOptions(options);
+  const result = await visualApi.mergeBaselines({
+    projectName: options.project,
+    ...options,
+  });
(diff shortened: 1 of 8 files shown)


visual-js/visual/src/cli/baselines.ts
```diff
@@ -0,0 +1,56 @@
...
+
+function apiFromOptions(opts: { region: VisualApiRegion }): VisualApi {
+  return getApi({
+    region: opts.region,
+    user: process.env.SAUCE_USERNAME,
+    key: process.env.SAUCE_ACCESS_KEY,
+  });
+}
+
+type BuildStatusCommandOptions = {
```
FriggaHel, line 13: I guess that should be this:

```suggestion
type BaselineMergeCommandOptions = {
```

visual-js/visual/src/cli/baselines.ts
```diff
@@ -0,0 +1,56 @@
...
+  project: string;
+  sourceBranch: string;
+  targetBranch: string;
+  region: VisualApiRegion;
+};
+export async function baselineMergeCommand(_: unknown, cmd: Command) {
+  const options = cmd.optsWithGlobals<BuildStatusCommandOptions>();
```
FriggaHel, line 20:
```suggestion
  const options = cmd.optsWithGlobals< BaselineMergeCommandOptions >();
```

visual-js/visual/src/cli/build.ts
```diff
@@ -104,26 +88,23 @@ export const buildStatusCommand = async (
...
-  region?: VisualApiRegion;
+  region: VisualApiRegion;
 };
 const statusCommand = () =>
   new Command()
     .name('status')
     .description('Fetches status from a Sauce Visual build')
     .addOption(buildIdOption)
     .addOption(customIdOption)
-    .addOption(regionOption)
```
FriggaHel, line 115: Moving this out of `build.ts` will make the region unavailable as soon as it's placed after `build` sub command (like `npx visual build create --name test_build --region eu-central-1`).

That's a bit annoying but this is the way Commander works as I observed id.

When I previously worked on this, I found no way other that duplicating that `region` param in each sub-command. It would be great if you finally got a solution for this.

ebekebe: I think duplicating is fine. I'll do this

FriggaHel (approved):

PR #123, opened 2024-08-30
Enable merge baselines command

ebekebe:
> Issue : IRIS-1166

### DIFF
--- a/visual-js/visual/src/cli/baselines.ts
+++ b/visual-js/visual/src/cli/baselines.ts
@@ -1,6 +1,6 @@
 import { Command, CommandOptions } from 'commander';
 import { VisualApiRegion } from '../common/regions';
-import { VisualApi, getApi } from '../api';
+import { VisualApi, getApi } from '../common/api';
 import { regionOption } from './common-options';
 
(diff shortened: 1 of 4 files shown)


visual-js/visual/src/cli/baselines.ts
```diff
@@ -23,7 +23,8 @@ export async function baselineMergeCommand(_: unknown, cmd: Command) {
   const visualApi = apiFromOptions(options);
   const result = await visualApi.mergeBaselines({
     projectName: options.project,
-    ...options,
+    sourceBranch: options.sourceBranch,
+    targetBranch: options.targetBranch,
```
FriggaHel, line 27: Should we check that they are different ? (and that this is non null).

That would permit to produce cleaner errors

ebekebe: 1. The null check should be carried out by commander, since those are required options.
2. Target and source branch can be the same. In this case nothing is copied (since `created_at(target) < created_at(source)` is false for all baselines)

FriggaHel (approved):

PR #125, opened 2024-09-09
[Playwright] Add Playwright Support

omacranger:
## Description
Adds Playwright support for Sauce Visual.

## Types of Changes

- Refactor/improvements


visual-js/visual-playwright/src/api.ts
```diff
@@ -0,0 +1,325 @@
...
+              return selectorRegions;
+            },
+            { selectors },
+          );
+        }
+
+        ignoreRegions = [
+          ...userIgnoreRegions.filter(filterIgnoreRegion),
+          ...selectorRegions,
+        ].filter((region) => 0 < region.width * region.height);
```
FriggaHel, line 230: We should probably check that both width / height are positive.
Having 2 negative value will result in the region to pass that filter while it's invalid. Isn't it ?

visual-js/visual-playwright/src/api.ts
```diff
@@ -0,0 +1,325 @@
...
+          if (!clipElement) {
+            return undefined;
+          }
+
+          const clientDims = clipElement.getBoundingClientRect();
+          const { x, y, height, width } = clientDims;
+
+          // If any values are falsy (or zero), then do not clip as those are invalid options for
+          // playwright
+          if (x < 0 || y < 0 || !height || !width) {
```
FriggaHel, line 252: Here shouldn't we use `Math.max(x, 0)` when `x+height > 0` ? Same applies to `y` coordinates.

If clipSelector is partially out of viewport (with x-coord negative), shouldn't we crop the visible area ?

visual-js/visual-playwright/src/types.ts
```diff
@@ -0,0 +1,35 @@
...
+  /**
+   * Whether we should capture a dom snapshot.
+   */
+  captureDom?: boolean;
+  /**
+   * A custom selector to clip to.
+   */
+  clipSelector?: string;
+  /**
+   * A number, in ms, that we should delay the snapshot by. Useful if the beginning of the story
```
FriggaHel, line 15: I guess that comment should be updated as there is no story here.

visual-js/visual-playwright/src/utils.ts
```diff
@@ -0,0 +1,130 @@
...
+    case 'darwin':
+      return OperatingSystem.Macos;
+    case 'win32':
+    case 'cygwin':
+      return OperatingSystem.Windows;
+    case 'freebsd':
+    case 'linux':
+    case 'openbsd':
+    case 'netbsd':
+      return OperatingSystem.Linux;
```
FriggaHel, line 88: Shouldn't we move BSD to `null` ?
It's still fairly different from Linux.

Btw who will ever use pw on OpenBSD ? (I was not even aware this was supported)

visual-js/visual-playwright/src/api.ts
```diff
@@ -0,0 +1,325 @@
...
+export const sauceVisualCheck = async (
+  page: Page,
+  testInfo: TestInfo,
+  name: string,
+  options?: Partial<SauceVisualParams>,
+) =>
+  await takePlaywrightScreenshot(
+    page,
+    {
+      deviceName: testInfo.project.name,
```
FriggaHel, line 143: Is deviceName expected to contain project name ?

visual-js/visual/src/utils.ts
```diff
@@ -122,3 +126,101 @@ export const parseRegionsForAPI = async <T>(
...
+ * Dom capturing script path
+ */
+
+export const domScriptPath = `${os.tmpdir()}/sauce-visual-dom-capture.js`;
+
+/**
+ * Fetch and save to tmp file dom capturing script
+ */
+
+export const downloadDomScript = async (api: ReturnType<typeof getApi>) => {
```
FriggaHel, line 140: I guess there is noway we can keep that script in memory only ? (to be 100% sure there is no left-overs)

visual-js/visual/src/utils.ts
```diff
@@ -122,3 +126,101 @@ export const parseRegionsForAPI = async <T>(
...
+  } catch (err: unknown) {
+    if ((err as NodeJS.ErrnoException).code !== 'ENOENT') {
+      console.error(err);
+    }
+  }
+};
+
+export const removeDomScriptFile = async () => {
+  try {
+    await fs.unlink(domScriptPath);
```
FriggaHel, line 165: There will still be the folder created by `os.tmpdir()` present. We should probably remove it too.

visual-js/visual-playwright/package.json
```diff
@@ -0,0 +1,76 @@
...
+  "main": "build/index.cjs",
+  "module": "build/index.js",
+  "types": "build/index.d.ts",
+  "license": "MIT",
+  "files": [
+    "build"
+  ],
+  "type": "module",
+  "engines": {
+    "node": "^16.13 || >=18"
```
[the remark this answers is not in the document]
omacranger: Yeah, but maybe this would be better to do for all of the SDKs at once?

visual-js/visual-playwright/src/api.ts
```diff
@@ -0,0 +1,325 @@
...
+      retry: (e) => e !== buildNotFoundError,
+    });
+
+    console.info(
+      `Your Sauce Visual Build is ready for review:
+  ${build.url}
+`,
+    );
+
+    if (build.unapprovedCount > 0) {
```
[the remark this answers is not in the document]
omacranger: This is currently unchanged from the Storybook integration... But yes that would probably make sense.

visual-js/visual-playwright/src/api.ts
omacranger: Yeah, Playwright doesn't really expose 'devices', but instead [projects](https://playwright.dev/docs/test-projects) which can be used to create configurations for runs. They are labelled in the config so we would just apply those labels here since some may be emulated.

visual-js/visual-playwright/src/api.ts
```diff
@@ -0,0 +1,325 @@
...
+    deviceName: string | undefined;
+  },
+  name: string,
+  options?: Partial<SauceVisualParams>,
+) => {
+  const { testName, suiteName, deviceName } = info;
+  const { buildId } = getOpts();
+
+  if (!buildId) {
+    return;
```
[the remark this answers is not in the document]
omacranger: Also current behavior for Storybook, but I'll see if we'll get any useful output with a warning here.

visual-js/visual-playwright/src/api.ts
omacranger: Yeah, I think that would work.

visual-js/visual-playwright/src/utils.ts
omacranger: This is also copied from Storybook -- but is just handling all the strings provided by TS for `os.platform()`.

visual-js/visual/src/utils.ts
omacranger: _Maybe_ -- but this is designed to do it once for all of the workers during global setup / cleanup during teardown. Otherwise it would download a separate file for each process / worker.

visual-js/visual-playwright/package.json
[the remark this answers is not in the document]
FriggaHel: Since this one is brand new, I would not bother waiting for others TBH.

visual-js/visual/src/utils.ts
omacranger: That folder is persistent / the value returned globally by node. I don't think we need to remove that folder since we're only creating a file in it and other processes / items could also append items here.

visual-js/visual-playwright/src/api.ts
omacranger: Added a `Math.max` here for both.

FriggaHel (approved):

PR #137, opened 2024-09-12
[Playwright] Add integration tests

omacranger:


.github/workflows/playwright-integration.yml
```diff
@@ -0,0 +1,43 @@
+name: playwright (integration test)
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'visual-js/visual-playwright/**'
+      - .github/workflows/js-build.yml
```
FriggaHel, line 9: Is that a left-over or an actual prerequisite ? I guess the 2nd, but prefer to be sure.

visual-js/visual-playwright/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+WORKDIR app
+
+RUN corepack enable
+
+COPY . ./
+
+RUN yarn install && npm run build --workspaces --if-present
+
+COPY ./visual-playwright/integration-tests/ ./integration-tests/
```
FriggaHel, line 11: Isn't this useless ?

Hasn't `COPY . ./` already pushed that folder too ?

FriggaHel (approved):

.github/workflows/playwright-integration.yml
omacranger: Um, not a prerequisite per se, but changing of the JS Build system probably means we should run the integration tests for the affected tooling right?

visual-js/visual-playwright/Dockerfile
omacranger: True... But it pushes it into a nested folder. Let me see how it behaves if I remove that line.

.github/workflows/playwright-integration.yml
FriggaHel: True 👍

FriggaHel (approved):

PR #138, opened 2024-09-19
Espresso SDK

kb-kerem:
Open points:

- ~~Environment variable handling needs to be adapted to Android best practices~~ Handled it with buildConfig and localProperties
- ~~Builder classes where appropriate (VisualClient, VisualCheckOptions, BuildAttributes etc.)~~ Provided
- ~~Use default Android http client instead of Apache's~~ Switched to using apollo graphql client


visual-espresso/visual/src/main/java/com/saucelabs/visual/exception/VisualApiException.java
```diff
@@ -0,0 +1,12 @@
+package com.saucelabs.visual.exception;
+
+public class VisualApiException extends RuntimeException {
```
konraddysput, line 3: Runtime Exception is a special exception - are you sure you want to extend the Runtime exception?  Throwing such an exception should kill the test suite. I think we want the test to fail instead?
```suggestion
public class VisualApiException extends Exception {
```

visual-espresso/visual/src/main/java/com/saucelabs/visual/graphql/GraphQLClient.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    public <D> D execute(GraphQLOperation operation, Class<D> responseType)
+            throws VisualApiException {
+        try {
+            HttpPost request = new HttpPost(this.uri);
+            request.setHeader(HttpHeaders.CONTENT_TYPE, "application/json");
+            request.setHeader(HttpHeaders.ACCEPT, "application/json");
+            request.setHeader(HttpHeaders.AUTHORIZATION, "Basic " + authentication);
+            request.setHeader(HttpHeaders.USER_AGENT, "sauce-visual-espresso/0.0.1");
```
konraddysput, line 66: You can read SDK version via build config

visual-espresso/visual/src/main/java/com/saucelabs/visual/junit/TestMetaInfoRule.java
```diff
@@ -0,0 +1,28 @@
...
+                try {
+                    TestMetaInfo.THREAD_LOCAL.set(new TestMetaInfo(className, testName));
+                    base.evaluate();  // Run the test
+                } finally {
+                    TestMetaInfo.THREAD_LOCAL.remove();
+                }
+            }
+        };
+    }
+}
```
konraddysput, line 28: missing end line in the end of file

konraddysput (commented):
Can we please comments in the public API?

omacranger:
A lot of the code here looks similar-ish to the Java client. Can none of that be imported and used here as a dependency somehow?

visual-espresso/visual/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -0,0 +1,140 @@
...
+public class VisualApi {
+    private final static String LOG_TAG = VisualApi.class.getSimpleName();
+
+    private final GraphQLClient graphQLClient;
+
+    public VisualApi(GraphQLClient graphQLClient) {
+        this.graphQLClient = graphQLClient;
+    }
+
+    VisualBuild createBuild(BuildAttributes buildAttributes) {
```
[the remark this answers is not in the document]
kb-kerem: I understand your concern but leaving out the accessor in java has a special meaning and cannot be expressed with another accessor. And having no accessor here is the most convenient accessor here in this case.
See: https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html

visual-espresso/visual/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -0,0 +1,140 @@
...
+        try {
+            byte[] screenshot = getScreenshot();
+            uploadToUrl(data.result.imageUploadUrl, screenshot);
+        } catch (IOException e) {
+            throw new VisualApiException("uploadSnapshot failed");
+        }
+        return data;
+    }
+
+    private byte[] getScreenshot() throws IOException {
```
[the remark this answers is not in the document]
kb-kerem: 1. makes sense, will do.
2. `os.close()` can cause the IOException

visual-espresso/visual/src/main/java/com/saucelabs/visual/VisualBuild.java
```diff
@@ -0,0 +1,51 @@
...
+        return id;
+    }
+
+    private static volatile VisualBuild build = null;
+
+    public static VisualBuild getBuildOnce(VisualApi visualApi, BuildAttributes buildAttributes) {
+        synchronized (VisualBuild.class) {
+            if (build == null) {
+                build = visualApi.createBuild(buildAttributes);
+                Runtime.getRuntime().addShutdownHook(new Thread(() -> visualApi.finishBuild(build.id)));
```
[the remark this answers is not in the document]
kb-kerem: The shutdownHook thing doesn't work in Android and this line has been removed altogether

visual-espresso/visual/src/main/java/com/saucelabs/visual/junit/TestMetaInfoRule.java
kb-kerem: added

kb-kerem:
> A lot of the code here looks similar-ish to the Java client. Can none of that be imported and used here as a dependency somehow?

@omacranger That would be super convenient however it brings also some problems that complicates the implementation. Importing Java client brings transitive dependencies that are not compatible with Android environment, considering the target Android API (apache httpclient, graphql-java etc.). Those dependencies need to be excluded explicitly and need to be maintained for the future versions of java client if new ones show up. Also, since we are using the same package and class names for certain classes that need reimplementation (mainly VisualAPI and GraphQLClient), class/method loading problems arise in runtime. Instead of trying to fix those problems, I'd rather keep the implementations fully separate.

visual-espresso/visual/src/main/java/com/saucelabs/visual/espresso/VisualClient.java
```diff
@@ -0,0 +1,91 @@
...
+    public void sauceVisualCheck(String snapshotName, VisualCheckOptions options) {
+        CreateSnapshotUploadMutation.Data data = visualApi.uploadSnapshot(this.build.getId());
+        CreateSnapshotMutation.SnapshotIn input = new CreateSnapshotMutation.SnapshotIn(
+                this.build.getId(),
+                data.result.id,
+                snapshotName,
+                options.resolveTestName(),
+                options.resolveSuiteName(),
+                OperatingSystem.ANDROID,
+                "");
```
omacranger, line 84: Are we able to get OS version here?

visual-espresso/visual/build.gradle
```diff
@@ -0,0 +1,77 @@
+plugins {
+    alias(libs.plugins.android.library)
+    id 'maven-publish'
+}
+
+android {
+    namespace 'com.saucelabs.visual'
```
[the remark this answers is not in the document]
kb-kerem: We do not need to change this namespace. Although this looks like a java package name, they're not coupled or related.

graphql-java unfortunately requires minimum Android API level 26 and we're targeting 21. I'll give this one a try and see if it helps: https://www.apollographql.com/docs/kotlin/

visual-espresso/visual/src/main/java/com/saucelabs/visual/espresso/VisualClient.java
kb-kerem: yes, added

konraddysput (approved):

visual-espresso/visual/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -0,0 +1,140 @@
...
+            os.write(file);
+            os.flush();
+
+            // Get the response code
+            int responseCode = connection.getResponseCode();
+            if (responseCode != HttpURLConnection.HTTP_OK) {
+                throw new VisualApiException("Snapshot upload failed. Response code: " + responseCode);
+            }
+        } catch (Exception e) {
+            throw new VisualApiException("Snapshot upload failed", e);
```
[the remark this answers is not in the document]
kb-kerem: Reworked the exception handling, thanks for the tip

visual-espresso/visual/src/main/java/com/saucelabs/visual/exception/VisualApiException.java
kb-kerem: This doesn't kill the suite but fails the test instead. Same approach we took in java-client. Tested and works fine.

visual-espresso/visual/src/androidTest/java/com/saucelabs/visual/SmokeTest.java
```diff
@@ -0,0 +1,12 @@
+package com.saucelabs.visual;
+
+import org.junit.Test;
+
+public class SmokeTest {
+
+    @Test
+    public void check() {
+        VisualClient visual = new VisualClient("Espresso smoke test");
+        visual.sauceVisualCheck("snap");
```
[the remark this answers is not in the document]
kb-kerem: done

PR #139, opened 2024-09-24
[Cypress] Ensure all statuses are available in `sauceVisualResult` summary

omacranger:
## Description
Our [docs](https://docs.saucelabs.com/visual-testing/integrations/cypress/#test-results-summary) say you can use the sauceVisualResult summary to check the status of a result and assert that a certain status is never found. Currently, however, that object is empty so asserting `unapproved = 0` would never pass. 

This PR does the following:

- This updates the check to populate all the keys of our DiffStatus into the initial value before reducing over the nodes.
- This also bumps all other bindings by a patch to roll out the new server-side ignore region calculation.

## Types of Changes

### DIFF
--- a/visual-js/.changeset/slow-seas-occur.md
+++ b/visual-js/.changeset/slow-seas-occur.md
@@ -1,0 +1,10 @@
+---
+"@saucelabs/cypress-visual-plugin": patch
+"@saucelabs/visual": patch
+"@saucelabs/nightwatch-sauce-visual-service": patch
+"@saucelabs/visual-playwright": patch
+"@saucelabs/visual-storybook": patch
+"@saucelabs/wdio-sauce-visual-service": patch
+---
+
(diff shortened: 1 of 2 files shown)


paweltomaszewskisaucelabs (approved):

visual-js/.changeset/slow-seas-occur.md
```diff
@@ -0,0 +1,11 @@
+---
+"@saucelabs/cypress-visual-plugin": patch
+"@saucelabs/visual": patch
```
paweltomaszewskisaucelabs, line 3: isn't it only impacting @saucelabs/cypress-visual-plugin pck?

paweltomaszewskisaucelabs (changes requested):
Isn't this change impacting only

paweltomaszewskisaucelabs (approved):

PR #143, opened 2024-09-30
Espresso ignore regions

kb-kerem:


visual-espresso/visual/src/main/java/com/saucelabs/visual/espresso/VisualCheckOptions.java
```diff
@@ -1,17 +1,45 @@
...
+            this.testName = testName;
+            return this;
+        }
+
+        public Builder withSuiteName(String suiteName) {
+            this.suiteName = suiteName;
+            return this;
+        }
+
+        public Builder withIgnoreRegions(List<RegionIn> ignoreRegions) {
```
omacranger, line 34: Should we also have a method here for passing elements and generating these regions based on the locations / coordinates from those? Is this common practice in Espresso (to assign elements to a variable to assertions on them, etc)?

omacranger (approved):

visual-espresso/visual/src/main/java/com/saucelabs/visual/espresso/VisualCheckOptions.java
konraddysput: I think this might be a good idea to allow the user to pass an element and handle the coordinates by ourself

visual-espresso/visual/src/main/java/com/saucelabs/visual/espresso/VisualCheckOptions.java
```diff
@@ -1,17 +1,45 @@
...
+            this.ignoreRegions = ignoreRegions;
```
konraddysput, line 35: idea:

since this is a builder I think, someone can try to pass ignore regions multiple times for different ignore regions based on the domain of the test


.withIgnoreRegions(ignoreRegionsForHeaderTime)
.withIgnoreRegions(ignoreRegionsForFooterContact)
.withIgnoreRegions(allDates)
.withIgnoreRegions(allMockNames)

Maybe we should consider list concatenation instead?

visual-espresso/visual/src/main/java/com/saucelabs/visual/espresso/VisualClient.java
```diff
@@ -92,14 +92,14 @@ public VisualClient build() {
...
-        return this.sauceVisualCheck(snapshotName, new VisualCheckOptions());
+        return this.sauceVisualCheck(snapshotName, new VisualCheckOptions.Builder().build());
     }
 
     /**
      * Uploads and creates a snapshot with a given snapshotName and options
      *
      * @param snapshotName A name for the snapshot
-     * @param options Options for the VisualCheck
+     * @param options      Options for the VisualCheck
```
konraddysput, line 102: is this expected?

kb-kerem: Yes, just standard javadoc formatting

PR #146, opened 2024-10-03
Espresso capture dom

kb-kerem:


visual-espresso/visual/src/main/java/com/saucelabs/visual/espresso/VisualClient.java
```diff
@@ -102,7 +113,8 @@ public CreateSnapshotMutation.Data sauceVisualCheck(String snapshotName) {
      * @param options      Options for the VisualCheck
      */
     public CreateSnapshotMutation.Data sauceVisualCheck(String snapshotName, VisualCheckOptions options) {
-        CreateSnapshotUploadMutation.Data data = visualApi.uploadSnapshot(this.build.getId());
+        Boolean captureDom = Optional.ofNullable(this.captureDom).orElse(options.getCaptureDom());
```
omacranger, line 116: These should perhaps be switched? Use the options first if present, then fallback to the global configuration?

omacranger (approved):

PR #148, opened 2024-10-08
Espresso selective diffing

kb-kerem:


visual-espresso/visual/src/main/java/com/saucelabs/visual/VisualCheckOptions.java
```diff
@@ -1,81 +1,101 @@
...
-        public Builder withIgnoreRegions(View... regions) {
-            List<RegionIn> ignoreRegions = new ArrayList<>();
-            for (View view : regions) {
-                ignoreRegions.add(GetRegionAction.toRegion(view));
+        public Builder captureDom(boolean captureDom) {
+            this.captureDom = captureDom;
+            return this;
+        }
+
+        public Builder regions(SelectiveRegion... selectiveRegions) {
```
omacranger, line 81: With the backend diffing support not being available for these yet, maybe we should hold off on exposing this as a part of a public API?

kb-kerem: Yes that makes sense. I'll leave out capture dom as well

kb-kerem: actually captureDom is possible but we're not able to calculate the dom diff atm. I'll leave that one in the public API.

omacranger (approved):

PR #152, opened 2024-10-11
[JS] Add ability to assert against visual results, Add upload retry

omacranger:
## Description

Adds the ability to assert against Sauce Visual results during test executions on Playwright and unifies the code for other frameworks already utilizing it.

Can either use the `sauceVisual.visualResults()` fixture or `sauceVisualResults` function directly to get the result summary for an individual test and make assertions on it like what is available in Cypress / WDIO.

Ex: 

```ts
test('should be able to login with standard user', async ({ page, sauceVisual }) => {
       // .. other test commands here
        await sauceVisual.visualCheck("Inventory Page");
        expect(await sauceVisual.sauceVisualResults()).toHaveProperty(DiffStatus.Unapproved, 0);
});
```

This also adds a retry mechanism for DOM / Snapshot uploads to GCS.

## Types of Changes


visual-js/visual-playwright/src/fixtures.ts
```diff
@@ -1,11 +1,23 @@
...
 import { SauceVisualParams } from './types';
-import { sauceVisualCheck } from './playwright';
+import { sauceVisualCheck, sauceVisualResults } from './playwright';
+import { DiffStatus } from '@saucelabs/visual';
 
 export type SauceVisualFixtures = {
   sauceVisual: {
+    /**
+     * Takes a snapshot of the current page and uploads it to Sauce Labs for visual diffing.
+     * @param name
```
konraddysput, line 11: can we add a descriptions for each parameters?

visual-js/visual-playwright/src/api.ts
```diff
@@ -318,12 +329,64 @@ ${e instanceof Error ? e.message : JSON.stringify(e)}
...
+        return diffsForTestResult.nodes
+          .filter(filterDiffsById)
+          .reduce((statusSummary, diff) => {
+            statusSummary[diff.status]++;
+            return statusSummary;
+          }, summary);
+      },
+      {
+        maxDelay: 10000,
+        numOfAttempts: 10,
```
konraddysput, line 387: how did we pick those numbers?

konraddysput (approved):
Just in case, I think we should get additional check from @ebekebe or @paweltomaszewskisaucelabs

ebekebe (approved):
LGTM

visual-js/visual-playwright/src/api.ts
omacranger: By rolling a die.

They're just general numbers to not spend too long on a single check, but allow retries. I think in the future it may be good to unify all retry values into a single options object so we can stick with something predefined.

visual-js/visual-playwright/src/api.ts
```diff
@@ -318,12 +329,64 @@ ${e instanceof Error ? e.message : JSON.stringify(e)}
...
       suiteName,
       uploadUuid: uploadId,
     });
+
+    if (!this.uploadedDiffIds[testId]) this.uploadedDiffIds[testId] = [];
+
+    this.uploadedDiffIds[testId].push(...diffs.nodes.map((diff) => diff.id));
+  }
+
+  public async visualResults({ testId }: { testId: string }) {
```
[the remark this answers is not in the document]
omacranger: Agreed. IMO I'd like to keep it like it is now since some of the behavior for grouping / ID checking is different than the other implementation, but I'll create a ticket to unify it for all the JS frameworks.

visual-js/visual-playwright/src/fixtures.ts
omacranger: Do `name` and `options` really need descriptions? 😅 Those seem pretty obvious.

ebekebe (approved):

PR #159, opened 2024-10-30
FPS native java

kb-kerem:


visual-java/src/main/java/com/saucelabs/visual/model/FullPageScreenshotConfig.java
```diff
@@ -56,13 +62,19 @@ public Builder withScrollLimit(int scrollLimit) {
       return this;
     }
 
+    public Builder withScrollElement(WebElement scrollElement) {
```
konraddysput, line 65: worth adding a note that this is being used only on native mobile

visual-java/src/main/java/com/saucelabs/visual/model/FullPageScreenshotConfig.java
```diff
@@ -105,4 +117,12 @@ public int getScrollLimit() {
   public void setScrollLimit(int scrollLimit) {
     this.scrollLimit = scrollLimit;
   }
+
+  public String getScrollElement() {
+    return scrollElement;
+  }
+
+  public void setScrollElement(WebElement scrollElement) {
```
konraddysput, line 125: worth adding a note that we're using it only for native mobile

visual-java/src/main/java/com/saucelabs/visual/model/FullPageScreenshotConfig.java
```diff
@@ -105,4 +117,12 @@ public int getScrollLimit() {
   public void setScrollLimit(int scrollLimit) {
     this.scrollLimit = scrollLimit;
   }
+
+  public String getScrollElement() {
```
konraddysput, line 121: I think it would be better if we rename the method to getScrollElementId

konraddysput (approved):

visual-java/src/main/java/com/saucelabs/visual/model/FullPageScreenshotConfig.java
kb-kerem: it's used by json builder so it has to match the name in spec

PR #168, opened 2024-11-27
[Storybook] Add ability to take snapshots during Storybook `play` interactions

omacranger:
## Description
This PR adds the ability to take snapshots _during_ Storybook play interactions, instead of just at the end. This also reintroduces Storybook 6 support by tweaking dependency requirements.

For example, you could now open a dropdown, focus an input field, type text to generate form errors all while taking snapshots at each step of the way to catch all visual edge cases without creating individual stories for each.

This requires a new `preVisit` hook which has to be added to the Storybook test runner configuration. If attempting to use the `takeVisualSnapshot` without it, it will throw an error and mention this new addition.

## Types of Changes

- Configuration change

### DIFF
--- a/visual-js/visual-storybook/src/api.ts
+++ b/visual-js/visual-storybook/src/api.ts
@@ -146,6 +146,13 @@ export const postVisit = async (page: Page, context: TestContext) => {
  */
 export const postRender = postVisit;
 
+/**
+ * Playwright throws an exception if attempting to expose the same binding twice and does not
+ * expose a way for us to see if something has already been bound. Since we're only given access
+ * to the Page object during postVisit (not during setup) we can't ensure that it's only added once.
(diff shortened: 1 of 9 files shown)


visual-js/visual-storybook/src/api.ts
```diff
@@ -143,3 +145,32 @@ export const postVisit = async (page: Page, context: TestContext) => {
  * `postVisit` exported from this package instead.
  */
 export const postRender = postVisit;
+
+let hasExposed = false;
```
ebekebe, line 149: Could you add a comment about the intent and lifecycle of this?

omacranger: Added

ebekebe (approved):

PR #188, opened 2025-02-14
Initial visual-snapshots package

aga-szczepanska:
Generate a new package called `visual-snapshots` with `pdf` command (which for now does nothing).


visual-js/visual-snapshots/jest.config.cjs
```diff
@@ -0,0 +1,20 @@
...
+    // '^.+\\.[tj]sx?$' to process js/ts with `ts-jest`
+    // '^.+\\.m?[tj]sx?$' to process js/ts/mjs/mts with `ts-jest`
+    '^.+\\.tsx?$': [
+      'ts-jest',
+      {
+        useESM: true,
+      },
+    ],
+  },
+  setupFiles: ['<rootDir>/setEnvVars.ts'],
```
perf2711, line 19: The file has a bit cryptic file name when looking at it outside of Jest's context. Can we rename it to something like `jest.setup.ts`, or any name that would signify that it is used solely for tests?

visual-js/visual-snapshots/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "noExternal": [
+      "@apollo/client"
+    ]
+  },
+  "devDependencies": {
+    "@types/jest": "29.5.14",
+    "eslint": "8.57.1",
+    "eslint-config-prettier": "^8.3.0",
+    "eslint-plugin-prettier": "5.0.0",
+    "globals": "15.15.0",
```
perf2711, line 60: Is this package used anywhere?

visual-js/visual-snapshots/setEnvVars.ts
```diff
@@ -0,0 +1 @@
+process.env.EXISTING_ENV_VARIABLE= 'existing-env-variable-value';
```
perf2711, line 1: It seems like you're using this only in `params.spec.ts`. Why not add this to a `beforeAll` hook?

visual-js/visual-snapshots/src/commands/params.ts
```diff
@@ -0,0 +1,19 @@
+export const getParam = (
+  options: { [key: string]: string },
```
konraddysput, line 2:
```suggestion
  options: Record<string, string>,
```

visual-js/visual-snapshots/src/commands/params.ts
```diff
@@ -0,0 +1,19 @@
...
+  if (process.env[envVariableName]) {
+    return process.env[envVariableName];
+  }
+  if (defaultValue) {
+    return defaultValue;
+  }
+  throw new Error(
+    `Missing required argument. Please provide ${optionName} option, or set ${envVariableName} environment variable.`,
+  );
+};
```
perf2711, line 19: Do you intend to use this for getting options from either the command options passed from CLI or env variables? If so, maybe a better option would be to pass defaults into the `Option`s themselves?

https://github.com/tj/commander.js?tab=readme-ov-file#default-option-value

konraddysput (changes requested):

visual-js/visual-snapshots/tsconfig.json
```diff
@@ -0,0 +1,111 @@
...
+    // "allowArbitraryExtensions": true,                 /* Enable importing files with any extension, provided a declaration file is present. */
+    // "noResolve": true,                                /* Disallow 'import's, 'require's or '<reference>'s from expanding the number of files TypeScript should add to a project. */
+
+    /* JavaScript Support */
+    // "allowJs": true,                                  /* Allow JavaScript files to be a part of your program. Use the 'checkJS' option to get errors from these files. */
+    // "checkJs": true,                                  /* Enable error reporting in type-checked JavaScript files. */
+    // "maxNodeModuleJsDepth": 1,                        /* Specify the maximum folder depth used for checking JavaScript files from 'node_modules'. Only applicable with 'allowJs'. */
+
+    /* Emit */
+    "declaration": true,                              /* Generate .d.ts files from TypeScript and JavaScript files in your project. */
```
perf2711, line 54: Do we need declaration files for a CLI?

visual-js/visual-snapshots/tsconfig.json
```diff
@@ -0,0 +1,111 @@
...
+    // "jsxFactory": "",                                 /* Specify the JSX factory function used when targeting React JSX emit, e.g. 'React.createElement' or 'h'. */
+    // "jsxFragmentFactory": "",                         /* Specify the JSX Fragment reference used for fragments when targeting React JSX emit e.g. 'React.Fragment' or 'Fragment'. */
+    // "jsxImportSource": "",                            /* Specify module specifier used to import the JSX factory functions when using 'jsx: react-jsx*'. */
+    // "reactNamespace": "",                             /* Specify the object invoked for 'createElement'. This only applies when targeting 'react' JSX emit. */
+    // "noLib": true,                                    /* Disable including any library files, including the default lib.d.ts. */
+    // "useDefineForClassFields": true,                  /* Emit ECMAScript-standard-compliant class fields. */
+    // "moduleDetection": "auto",                        /* Control what method is used to detect module-format JS files. */
+
+    /* Modules */
+    "module": "commonjs",                                /* Specify what module code is generated. */
```
perf2711, line 28: `type` is set to `"module"` in `package.json`, which tells me that this should be an ESM project. Did you intend to compile the code also to ESM? 

To make Typescript also compile this as ESM, you need to make the following changes to `tsconfig.json`:

```json
{
  "target" : "ESNext",
  "module": "NodeNext",
  "moduleResolution": "nodenext"
}
```

All imports also have to be imported with a `.js` extension, so for example:
```ts
import { pdfCommand } from "./commands/pdf";
```
would have to be changed to:
```ts
import { pdfCommand } from "./commands/pdf.js";
```

visual-js/visual-snapshots/tsconfig.json
```diff
@@ -0,0 +1,111 @@
...
+    // "noImplicitOverride": true,                       /* Ensure overriding members in derived classes are marked with an override modifier. */
+    // "noPropertyAccessFromIndexSignature": true,       /* Enforces using indexed accessors for keys declared using an indexed type. */
+    // "allowUnusedLabels": true,                        /* Disable error reporting for unused labels. */
+    // "allowUnreachableCode": true,                     /* Disable error reporting for unreachable code. */
+
+    /* Completeness */
+    // "skipDefaultLibCheck": true,                      /* Skip type checking .d.ts files that are included with TypeScript. */
+    "skipLibCheck": true                                 /* Skip type checking all .d.ts files. */
+  }
+}
```
perf2711, line 111: Small nit: can we remove all the unused commented options? It'll make the config file easier to read.

visual-js/visual-snapshots/package.json
```diff
@@ -0,0 +1,65 @@
...
+    ],
+    "dts": true,
+    "outDir": "./build",
+    "format": [
+      "cjs",
+      "esm"
+    ],
+    "noExternal": [
+      "@apollo/client"
+    ]
```
perf2711, line 53: I'm not sure, is this necessary here?

visual-js/visual-snapshots/package.json
```diff
@@ -0,0 +1,65 @@
...
+  "license": "MIT",
+  "files": [
+    "build",
+    "README.md"
+  ],
+  "type": "module",
+  "engines": {
+    "node": "^16.13 || >=18"
+  },
+  "typeScriptVersion": "3.8.3",
```
perf2711, line 17: This seems to not correlate with the actual version of Typescript used in this project. Is this field necessary at all?

visual-js/visual-snapshots/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "dts": true,
+    "outDir": "./build",
+    "format": [
+      "cjs",
+      "esm"
+    ],
+    "noExternal": [
+      "@apollo/client"
+    ]
+  },
```
perf2711, line 54: Do we need `tsup` at all? If this will be a CLI tool, I think building by `tsc` will suffice.

visual-js/visual-snapshots/package.json
```diff
@@ -0,0 +1,65 @@
+{
+  "name": "@saucelabs/visual-snapshots",
+  "description": "CLI which generates Visual snapshots for a specified document",
+  "version": "0.13.0",
```
perf2711, line 4: Do you think starting from `0.0.1` would be better?

visual-js/visual-snapshots/README.md
```diff
@@ -0,0 +1,29 @@
...
+Execute:
+
+```sh
+node build/index.js [command] [params] 
+```
+
+Run tests:
+
+```sh
+npx jest
```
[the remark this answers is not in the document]
perf2711: Not really - `jest` will work inside of a script. So this will suffice:
```
// package.json
"scripts": {
  "test": "jest"
}
```

and then

```
npm run test
```

visual-js/visual-snapshots/setEnvVars.ts
aga-szczepanska: The file is redundant after setting the env in the test file

visual-js/visual-snapshots/jest.config.cjs
aga-szczepanska: The file has been deleted

visual-js/visual-snapshots/src/commands/params.ts
aga-szczepanska: We need to get the option value, or env variable value, or default value.
I could use the default and set to to env variable value or the default value... but not every option has a default value.
In my opinion, this is clearer this way and in case the value is not determined but required we throw an exception in a single place

visual-js/visual-snapshots/package.json
```diff
@@ -0,0 +1,65 @@
...
+  "main": "build/index.cjs",
+  "module": "build/index.js",
+  "types": "build/index.d.ts",
```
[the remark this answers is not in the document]
aga-szczepanska: with `tsup` it neede to be `build`, now that I switched to `tsc` it's build :)

visual-js/visual-snapshots/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "outDir": "./build",
+    "format": [
+      "cjs",
+      "esm"
+    ],
+    "noExternal": [
+      "@apollo/client"
+    ]
+  },
+  "devDependencies": {
```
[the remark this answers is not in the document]
aga-szczepanska: It's not used anymore

visual-js/visual-snapshots/tsconfig.json
aga-szczepanska: Done :)

visual-js/visual-snapshots/package.json
aga-szczepanska: Deleted

visual-js/visual-snapshots/package.json
aga-szczepanska: Switched to `tsc`

visual-js/visual-snapshots/package.json
```diff
@@ -0,0 +1,65 @@
+{
```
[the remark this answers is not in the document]
aga-szczepanska: Done

visual-js/visual-snapshots/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "name": "@saucelabs/visual-snapshots",
+  "description": "CLI which generates Visual snapshots from a data source such as pdf",
+  "version": "0.1.0",
+  "main": "lib/index.cjs",
+  "module": "lib/index.js",
```
perf2711, line 6: If you're using `tsc`, most likely you do not compile to `cjs` and `js`, but only to `js`. Can we change `main` to `"lib/index.js"` and remove `module`?

visual-js/visual-snapshots/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "types": "lib/index.d.ts",
```
perf2711, line 7: You're not emitting declaration files, so this can be removed.

visual-js/visual-snapshots/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "license": "MIT",
+  "bin": "./lib/index.js",
+  "files": [
+    "build",
```
perf2711, line 11: Can we change this to `lib`?

visual-js/visual-snapshots/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "pdf"
+  ],
+  "exports": {
+    ".": {
+      "require": "./lib/index.cjs",
+      "import": "./lib/index.js",
+      "types": "./lib/index.d.ts"
+    },
+    "./package.json": "./package.json"
+  },
```
perf2711, line 31: I think this is redundant and we can remove this. 
This is useful for multiple entry points, which we do not use in this app.

https://docs.npmjs.com/cli/v11/configuring-npm/package-json#exports

visual-js/visual-snapshots/src/commands/pdf.ts
```diff
@@ -0,0 +1,33 @@
...
+    .addOption(usernameOption)
+    .addOption(accessKeyOption)
+    .addOption(regionOption)
+    .addOption(buildNameOption)
+    .addOption(branchOption)
+    .addOption(defaultBranchOption)
+    .addOption(projectOption)
+    .addOption(buildIdOption)
+    .addOption(customIdOption)
+    .action((pdfFilePath: string, options: Record<string, string>) => {
```
perf2711, line 28: As we know which options to expect, maybe it would be a good idea to define an interface with the options?

```ts
interface PdfCommandOptions {
  username: string;
  accessKey: string;
  customId?: string;
  // etc
}
```

and then annotate options as:
```ts
options: PdfCommandOptions
```

aga-szczepanska: I have already created an interface when playing with visual a little and have it stashed. I don't think this needs to be a part of this MR.

visual-js/visual-snapshots/src/commands/pdf.ts
```diff
@@ -0,0 +1,33 @@
...
+  buildNameOption,
+  customIdOption,
+  defaultBranchOption,
+  projectOption,
+  regionOption,
+  usernameOption,
+} from "./options.js";
+
+export const pdfCommand = () => {
+  return new Command()
```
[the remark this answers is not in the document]
omacranger: Should be automatic: https://github.com/tj/commander.js?tab=readme-ov-file#automated-help

omacranger (approved):

visual-js/visual-snapshots/src/commands/pdf.ts
```diff
@@ -0,0 +1,33 @@
...
+    .name("pdf")
+    .description("Create visual snapshots for each page of a PDF file")
+    .argument("<pdf-file-path>", "A path to a PDF file")
```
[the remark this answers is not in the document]
konraddysput: sounds like a really good idea!

visual-js/visual-snapshots/src/commands/pdf.ts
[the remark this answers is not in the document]
aga-szczepanska: Yes, it works out of the box

visual-js/visual-snapshots/src/commands/pdf.ts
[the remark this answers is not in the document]
aga-szczepanska: I don't know how to do that now... but hopefully it will become clearer when I start actually dealing with Visual. I will bear it in mind :)

PR #198, opened 2025-03-11
visual-js: fix visual-storybook lint errors

perf2711:
# Fix `visual-storybook` lint errors

## Description
When running `npm run storybook` in `visual-storybook/integration-tests`, lint errors appear. Additionally, ESLint throws errors about invalid configuration (not all files are included by `tsconfig.json`).

## Types of Changes

### DIFF
--- a/visual-js/visual-storybook/integration-tests/tsconfig.json
+++ b/visual-js/visual-storybook/integration-tests/tsconfig.json
@@ -21,7 +21,6 @@
     "jsx": "react-jsx"
   },
   "include": [
-    "src/**/*.ts",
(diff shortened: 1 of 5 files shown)


visual-js/visual-storybook/integration-tests/tsconfig.json
```diff
@@ -21,6 +21,7 @@
     "jsx": "react-jsx"
   },
   "include": [
-    "src"
+    "src/**/*.ts",
```
omacranger, line 24: Shouldn't need to specify file types here `src/**/*` should be enough and it will know to choose ts/tsx.

PR #203, opened 2025-03-12
[INT-1] visual-snapshots: add named binary and update README

perf2711:
# `visual-snapshots`: add named binary and update README

> Issue : INT-1

## Description
Changes the binary name to `visual-snapshots` and updates the README.md file. 

## Types of Changes
- Refactor/improvements

### DIFF
--- a/visual-js/visual-snapshots/README.md
+++ b/visual-js/visual-snapshots/README.md
@@ -8,10 +8,15 @@ This package provides a CLI tool to create Visual snapshots of a provided PDF do
 node >= 18
 ```
 
-## Installation
+## Installation and usage
 
 ```sh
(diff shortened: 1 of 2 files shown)


omacranger:
Does this change interfere with calling it like our other Visual package and keeping it simple / less verbose? Ex:

`npx @saucelabs/visual-snapshots <command> [params]`

perf2711:
> Does this change interfere with calling it like our other Visual package and keeping it simple / less verbose? Ex:
> 
> `npx @saucelabs/visual-snapshots <command> [params]`

Yes, I believe this should still work.

perf2711:
Closing for now. `npx` is sufficient for this package.

PR #209, opened 2025-03-24
[no-ticket] always finish created builds

aga-szczepanska:
If snapshot creation fails, we should finish the created build

### DIFF
--- a/visual-js/visual-snapshots/src/app/pdf-handler.ts
+++ b/visual-js/visual-snapshots/src/app/pdf-handler.ts
@@ -12,11 +12,6 @@ export interface PdfCommandParams
   concurrency: number;
 }
 
-export enum PdfCommandStatus {
-  SUCCESS,
-  FAILURE,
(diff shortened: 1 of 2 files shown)


visual-js/visual-snapshots/src/app/pdf-handler.ts
```diff
@@ -12,28 +12,43 @@ export interface PdfCommandParams
...
+    try {
+      await this.pdfSnapshotUploader.uploadSnapshots({
+        buildId,
+        pdfFilePaths,
+        suiteName: params.suiteName,
+        testNameFormat: params.testName,
+        snapshotNameFormat: params.snapshotName,
+      });
+      status = PdfCommandStatus.SUCCESS;
+    } catch (_) {
```
omacranger, line 45: This could potentially swallow a number of errors and make it difficult to debug. Maybe we should have the error / message surfaced here at the least?

PR #210, opened 2025-03-24
[no-ticket] fix options

aga-szczepanska:
Pass all required


visual-js/visual-snapshots/src/app/worker/worker.ts
```diff
@@ -16,7 +25,16 @@ import { VisualSnapshotsApi } from "../../api/visual-snapshots-api.js";
...
+  .addOption(regionOption)
+  .addOption(buildNameOption)
+  .addOption(branchOption)
+  .addOption(defaultBranchOption)
+  .addOption(projectOption)
+  .addOption(buildIdOption)
+  .addOption(customIdOption)
+  .addOption(suiteNameOption)
+  .addOption(testNameOption)
+  .addOption(snapshotNameOption);
```
perf2711, line 37: This shouldn't be necessary here. All snapshot options are passed to the function. Did something not work here?

visual-js/visual-snapshots/src/commands/options.ts
```diff
@@ -80,6 +80,19 @@ export const suiteNameOption = new Option(
...
+    EOL +
+    "Supports the following parameters: {filename}"
+).default("{filename}");
+
+export const snapshotNameOption = new Option(
+  "--snapshot-name <snapshot-name>",
+  "The name of the snapshot you would like to appear in the Sauce Visual dashboard." +
+    EOL +
+    " Supports the following parameters: {filename}, {page}"
+).default("{filename}-{page}");
```
perf2711, line 95: I left this in `pdf.ts` because `{page}` may not make sense outside of PDF. Do you think it's better to keep it here?

visual-js/visual-snapshots/src/app/worker/worker.ts
aga-szczepanska: Using any of those arguments resulted in a failure caused by an invalid argument. Please try it yourself. I tried with `--test-name` and `--suite-name` options.

visual-js/visual-snapshots/src/commands/options.ts
aga-szczepanska: I moved it because I needed to import it in worker and I didn't want to import it from here.

visual-js/visual-snapshots/src/app/worker/worker.ts
perf2711: Ah! Then maybe a good idea would be to add instead this?

```ts
  .allowUnknownOption(true)
```

aga-szczepanska: It works :)

perf2711 (approved):

PR #220, opened 2025-04-17
[Java][WDIO] Add tolerance settings support

omacranger:
## Description
Adds support for selecting a preset or adjusting the tolerance settings for the diffing engines for both the Java and WDIO SDKs.

## Types of Changes

### DIFF
--- a/visual-java/src/main/java/com/saucelabs/visual/CheckOptions.java
+++ b/visual-java/src/main/java/com/saucelabs/visual/CheckOptions.java
@@ -2,10 +2,7 @@ package com.saucelabs.visual;
 
 import com.saucelabs.visual.graphql.type.DiffingOptionsIn;
 import com.saucelabs.visual.graphql.type.IgnoreSelectorIn;
-import com.saucelabs.visual.model.DiffingFlag;
-import com.saucelabs.visual.model.FullPageScreenshotConfig;
-import com.saucelabs.visual.model.IgnoreRegion;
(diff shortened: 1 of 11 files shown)


visual-java/src/main/java/com/saucelabs/visual/CheckOptions.java
```diff
@@ -49,6 +46,24 @@ public CheckOptions(
...
+    this.captureDom = builder.captureDom;
+    this.clipSelector = builder.clipSelector;
+    this.clipElement = builder.clipElement;
+    this.fullPageScreenshotConfig = builder.fullPageScreenshotConfig;
+    this.diffingOptions = builder.diffingOptions;
+    this.hideScrollBars = builder.hideScrollBars;
+    this.diffingMethodSensitivity = builder.diffingMethodSensitivity;
+    this.diffingMethodTolerance = builder.diffingMethodTolerance;
+  }
+
```
perf2711, line 66: It seems you cannot create `CheckOptions` manually, without using builder, with `diffingMethodSensitivity`/`diffingMethodTolerance`. Is this intended?

perf2711 (approved):

visual-java/src/main/java/com/saucelabs/visual/CheckOptions.java
perf2711: I'm fine with only using builder for that or setters. I just thought that this was an omission.

PR #222, opened 2025-04-22
Diffing Tolerance Settings for Dotnet, Python, Remaining JS SDKs

omacranger:
This adds diffing tolerance settings to the remaining SDKs, including dotnet, Python, Cypress, Playwright, Nightwatch.

## Description
A few sentences describing the overall goals of the pull request's
commits.

## Types of Changes


visual-js/visual-cypress/src/types.ts
```diff
@@ -64,11 +74,21 @@ export type VisualCheckOptions = {
   /**
    * The diffing method we should use when finding visual changes. Defaults to DiffingMethod.Balanced
    */
-  diffingMethod?: DiffingMethod;
+  diffingMethod?: `${DiffingMethod}`;
```
omacranger, line 77: Note that this is a 'breaking' type change, but since this doesn't work currently I think this is safe to do. Since you cannot import `DiffingMethod` from `@saucelabs/visual` without it blowing up a test due to bundling of modules from tsup.

perf2711: Why not set it to ``DiffingMethod | `${DiffingMethod}` `` like in `ScreenshotMetadata`?

perf2711 (approved):
Some files are missing end of file newlines. Do you think you could add them?

visual-js/visual-cypress/src/types.ts
omacranger: `DiffingMethod` doesn't work (you can't import from `@saucelabs/visual` without it breaking) so I didn't want to suggest to TS / IDEs to do so.

PR #223, opened 2025-04-24
[INT-215] template improvements

aga-szczepanska:
We used to have the following templates:
- page
- filename (with extension)

This is not enough, we need the following:
- page - stays as is
- filename - without extension
- ext - file extension, for instance _.pdf_
- dirname - name of the parent directory the file is located in, it doesn't matter what the current directory is
- dirname relative - relative path from current directory to the file

On top of that, all templates should be available for snapshot, test, and suite name.


visual-js/visual-snapshots/src/app/worker/pdf-page-snapshot-uploader.ts
```diff
@@ -14,23 +16,24 @@ export class PdfPageSnapshotUploader {
...
     snapshotNameFormat: string | undefined
   ) {
     const pdfFile = await this.pdfFileLoader.loadPdfFile(pdfFilePath);
     const page = await pdfFile.getPage(pageNumber);
 
-    const filename = path.basename(pdfFile.path);
+    const pdfFileMetadata = buildFileMetadata(pdfFilePath, pageNumber);
     const testName = testNameFormat
-      ? formatString(testNameFormat, { filename })
+      ? formatStringWithFileMetadata(testNameFormat, pdfFileMetadata)
```
aga-szczepanska, line 28: Do we want to enable all templates for each snapshot, test, and suite name? I think this way we make it more flexible but I am fine with limiting it

omacranger: If providing in that context makes sense I think the more flexibility the better provided we have pretty good defaults so it doesn't require someone to look through all the docs. I know that part is a separate ticket though.

omacranger (approved):

PR #225, opened 2025-04-25
[INT-201] reasonable default values

aga-szczepanska:
Create defaults for: build name, suite name, test name, snapshot name

### DIFF
--- a/visual-js/visual-snapshots/src/commands/options.ts
+++ b/visual-js/visual-snapshots/src/commands/options.ts
@@ -34,7 +34,9 @@ export const buildNameOption = new Option(
   "The name you would like to appear in the Sauce Visual dashboard." +
     EOL +
     "If not provided, SAUCE_VISUAL_BUILD_NAME environment variable will be used."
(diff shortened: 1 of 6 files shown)


perf2711 (approved):

visual-js/visual-snapshots/src/commands/options.ts
```diff
@@ -80,21 +82,21 @@ export const suiteNameOption = new Option(
   "The name of the suite you would like to appear in the Sauce Visual dashboard." +
     EOL +
     "Supports the following parameters: {filename}, {ext}, {directory}, {directoryRelative}, {page}"
-);
+).default("{directory}");
```
perf2711, line 85: Should `{directory}` be the default value here? It can be a bit unexpected IMO. Especially, when you pass multiple PDFs from different directories.

I propose leaving this undefined, or to set this to something generic, like "PDF Visual Tests".

aga-szczepanska: As agreed with the team, we will use `directoryRelative` as the default

PR #226, opened 2025-04-25
[no-ticket] Do not create a build when there are no files to process

aga-szczepanska:
- Do not create a build when there are no files to process
- Log files to be processed
- Extra test
- Divide test helpers into files
- Add tests which check that we only look for pdf files

### DIFF
--- a/visual-js/visual-snapshots/src/app/pdf-handler.ts
+++ b/visual-js/visual-snapshots/src/app/pdf-handler.ts
@@ -3,8 +3,10 @@ import {
   VisualSnapshotsApi,
 } from "../api/visual-snapshots-api.js";
 import { VisualConfig } from "@saucelabs/visual";
-import { getFiles } from "../utils/glob.js";
 import { PdfSnapshotUploader } from "./pdf-files-snapshot-uploader.js";
+import { logger as defaultLogger } from "../logger.js";
+import { FileExtractor } from "../utils/glob.js";
+import { Logger } from "pino";
 
 export interface PdfCommandParams
   extends VisualConfig,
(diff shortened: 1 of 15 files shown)


visual-js/visual-snapshots/src/app/pdf-handler.ts
```diff
@@ -15,14 +17,30 @@ export interface PdfCommandParams
...
+    if (pdfFilePaths.length === 0) {
+      this.logger.warn("No PDF files found to process.");
+      return;
+    }
+
+    this.logger.info(
+      `Found ${pdfFilePaths.length} PDF file${
+        pdfFilePaths.length > 1 ? "s" : ""
+      } to process: ${pdfFilePaths.join(", ")}`
+    );
```
perf2711, line 43: Perhaps instead of interpolating variables into the string, you could move them to logger args?

```suggestion
    this.logger.info(
      `Found PDF file${
        pdfFilePaths.length > 1 ? "s" : ""
      } to process`, {
        count: pdfFilePaths.length,
        paths: pdfFilePaths
    });
```

visual-js/visual-snapshots/test/app/pdf-handler.spec.ts
```diff
@@ -55,17 +70,71 @@ describe("pdf-handler", () => {
...
+          globOrDirs: [path.join(__dirname(import.meta), "../files/1.pdf")],
+          dirGlob: "*.pdf",
+        },
+      ]);
+      expect(logged).toEqual([
+        {
+          level: 30,
+          msg: "Found 1 PDF file to process: /absolute/path/to/files/1.pdf",
+        },
+      ]);
```
perf2711, line 127: Do you think `expect(logged).toMatchSnapshot()` would be a better idea?

perf2711 (approved):
Only small nits.

visual-js/visual-snapshots/test/app/pdf-handler.spec.ts
aga-szczepanska: As this is just a single log, I think this is simpler this way but I am fine with using snapshots if that's your preference.

PR #227, opened 2025-04-28
[Java] Add local selenium grid support for basic snapshots

omacranger:
## Description
Adds local selenium grid support to our Java SDK. This should automatically determine if this is a sauce session (In most instances) and either run our capture tooling from our backend or the simplified solution from our SDK directly using Selenium APIs.

Currently this omits the following:

- Full page screenshot support
- Element based ignore regions
- element / page clipping
- dom capture (and therefore selective diffing)

## Types of Changes

### DIFF
--- a/visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
+++ b/visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
@@ -11,23 +11,30 @@ import com.saucelabs.visual.model.DiffingMethodTolerance;
 import com.saucelabs.visual.model.FullPageScreenshotConfig;
 import com.saucelabs.visual.model.IgnoreRegion;
 import com.saucelabs.visual.model.VisualRegion;
+import com.saucelabs.visual.utils.CapabilityUtils;
 import com.saucelabs.visual.utils.ConsoleColors;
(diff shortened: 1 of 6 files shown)


visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -283,6 +322,41 @@ public VisualApi(
...
+    }
+
+    CommandExecutor executor = this.driver.getCommandExecutor();
+    Object resolvedExecutor = null;
+    if (executor instanceof TracedCommandExecutor) {
+      try {
+        final Field field = TracedCommandExecutor.class.getDeclaredField("delegate");
+        field.setAccessible(true);
+        resolvedExecutor = field.get(executor);
+      } catch (NoSuchFieldException | IllegalAccessException ignored) {
```
aga-szczepanska, line 341: What do you think about logging these? If something goes wrong, it will be difficult to debug

omacranger: I _think_ we'd be fine without it? We wouldn't get client side logs anyway, but this is just to eager set that value for the defaults in Selenium. it will default to Sauce if we can't verify it through this method, and users that want to force local can use our configuration options to skip this automatic detection.

PR #228, opened 2025-05-05
[Java/Local] Add DOM capture 

omacranger:
Adds DOM capture / upload for local selenium clients.

## Types of Changes

### DIFF
--- a/visual-java/src/main/java/com/saucelabs/visual/graphql/GraphQLClient.java
+++ b/visual-java/src/main/java/com/saucelabs/visual/graphql/GraphQLClient.java
@@ -7,14 +7,17 @@ import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
 import com.saucelabs.visual.exception.VisualApiException;
 import com.saucelabs.visual.utils.ArtifactVersion;
 import java.io.IOException;
+import java.net.URI;
 import java.util.Base64;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpHeaders;
 import org.apache.http.HttpResponse;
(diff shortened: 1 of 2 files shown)


visual-java/src/main/java/com/saucelabs/visual/graphql/GraphQLClient.java
```diff
@@ -63,18 +69,52 @@ public GraphQLClient(
...
-   * Uploads an image/png file to Sauce's visual storage.
+   * Uploads a file to Sauce's visual storage.
    *
    * @param uri The URL retrieved from the createSnapshotUpload mutation.
-   * @param image An image byte array, retrieved from the selenium driver's getScreenshotAs()
-   *     method.
+   * @param content A byte array containing the contents to upload to Visual storage.
    */
-  public void upload(String uri, byte[] image) throws VisualApiException {
+  public void upload(String uri, byte[] content, String mimeType) throws VisualApiException {
```
perf2711, line 113: As this is a public API, maybe `mimeType` should be optional and set to `"image/png"` by default? Or, add an overload without `mimeType` and set it as deprecated?

perf2711 (approved):

visual-java/src/main/java/com/saucelabs/visual/graphql/GraphQLClient.java
omacranger: This only exists on top of the `java-local` feature branch which hasn't been released yet, and even then we anticipate limited usage. Not sure it's worth adding / maintaining the override for it

perf2711 (approved):

PR #229, opened 2025-05-08
[no-ticket] fix invalid value being passed to ignoreChangesFor

perf2711:
# Fix invalid value being passed to `ignoreChangesFor`

## Description
`getHeight()` is passed instead of `getY()` to the `y` parameter.

## Types of Changes


omacranger:
Probably handled in a separate PR, but since this is passing x/y values instead of an element is this broken when the page is scrolled in advance?

omacranger (approved):

perf2711:
Perhaps. I noticed this while implementing the local grid ignore regions. I didn't check if the scrolling is working correctly or not.

PR #230, opened 2025-05-08
[INT-251] Support ignore regions for local snapshots

perf2711:
# Support ignore regions for local snapshots

> Issue : [INT-251](https://saucedev.atlassian.net/browse/INT-251)

## Description
Add support for ignore regions for local snapshots.
Ignore region coordinates are offset by the current page scroll.

## Types of Changes

[INT-251]: https://saucedev.atlassian.net/browse/INT-251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
+++ b/visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
@@ -21,10 +21,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import org.apache.http.client.config.RequestConfig;
-import org.openqa.selenium.By;
-import org.openqa.selenium.Capabilities;
-import org.openqa.selenium.OutputType;
-import org.openqa.selenium.WebElement;
+import org.openqa.selenium.*;
 import org.openqa.selenium.remote.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -641,6 +638,12 @@ public class VisualApi {
 
     WindowScroll scroll = getWindowScroll();
     List<RegionIn> ignoreRegions = extractIgnoreList(options);
+
+    for (WebElement element : options.getIgnoreElements()) {
+      RegionIn ignoreRegion = VisualRegion.ignoreChangesFor(element).toRegionIn();
+      ignoreRegions.add(ignoreRegion);
+    }
+
     for (RegionIn region : ignoreRegions) {
       region.setX(region.getX() - scroll.x);
       region.setY(region.getY() - scroll.y);
--- a/visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
+++ b/visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
@@ -652,8 +652,8 @@ public class VisualApi {
     }
 
     for (RegionIn region : ignoreRegions) {
-      region.setX(region.getX() - scroll.x);
-      region.setY(region.getY() - scroll.y);
+      region.setX(region.getX() - scroll.getX());
+      region.setY(region.getY() - scroll.getY());
     }
 
     // create snapshot using upload id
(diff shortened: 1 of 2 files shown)


perf2711:
This contains the fix commits from other pull requests. As the `feat/java-local-dom` branch is updated to have these commits, or merged to `main`, the commits will be removed.

visual-java/src/main/java/com/saucelabs/visual/model/WindowScroll.java
```diff
@@ -0,0 +1,11 @@
+package com.saucelabs.visual.model;
+
+public class WindowScroll {
+  public int x;
```
aga-szczepanska, line 4: Could you please make these private and create getters? I think we only want to read them, right?

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -798,7 +821,7 @@ private List<RegionIn> extractIgnoreList(CheckOptions options) {
         options.getIgnoreRegions() == null ? Arrays.asList() : options.getIgnoreRegions();
 
     List<VisualRegion> visualRegions =
-        options.getIgnoreRegions() == null ? Arrays.asList() : options.getRegions();
+        options.getRegions() == null ? Arrays.asList() : options.getRegions();
```
aga-szczepanska, line 824: Isn't it the same change as in this PR? https://github.com/saucelabs/visual-sdks/pull/231

aga-szczepanska (approved):

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
perf2711: Yeah, this is the same, but I kept this here for now as without this, it's simply not working. The source branch is @omacranger's, so unless he rebases with `main`, I have to include this.

As said above:
>This contains the fix commits from other pull requests. As the feat/java-local-dom branch is updated to have these commits, or merged to main, the commits will be removed.

visual-java/src/main/java/com/saucelabs/visual/model/WindowScroll.java
perf2711: I changed them to be `final`. Wdyt?

perf2711:
Waiting for `feat/java-local-dom` to be merged to `main`.

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -642,6 +636,19 @@ private void sauceVisualCheckLocal(String snapshotName, CheckOptions options) {
...
+    WindowScroll scroll = getWindowScroll();
+    List<RegionIn> ignoreRegions = extractIgnoreList(options);
+
+    for (WebElement element : options.getIgnoreElements()) {
+      RegionIn ignoreRegion = VisualRegion.ignoreChangesFor(element).toRegionIn();
+      ignoreRegions.add(ignoreRegion);
+    }
+
+    for (RegionIn region : ignoreRegions) {
+      region.setX(region.getX() - scroll.x);
```
omacranger, line 648: This should be `region.getX() + scroll.x`, no? Though this may only apply when doing full page

visual-java/src/main/java/com/saucelabs/visual/model/WindowScroll.java
omacranger: I agree. Plus, we're already following the getter/setter pattern so it makes more sense to keep things aligned here.

perf2711: Changed to getters.

omacranger: Was this pushed? 🤔 I still see `.x` usage.

omacranger (approved):

PR #232, opened 2025-05-09
[INT-252] [Java/Local] add screenshot clipping

perf2711:
# [Java/Local] add screenshot clipping

> Issue : INT-252

## Description
Add screenshot clipping to selected clip element or clip selector.
The element has to be visible on the initial screenshot for clipping to succeed.

## Types of Changes


visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -729,6 +752,54 @@ private DiffingMethodTolerance getDiffingMethodTolerance(CheckOptions checkOptio
...
+
+    return null;
+  }
+
+  private byte[] cropImage(
+      byte[] imageData, String imageFormat, int x, int y, int width, int height) {
+    // We can check that before loading the image
+    if (x < 0 || y < 0) {
+      throw new VisualApiException("Clipped element is not visible.");
+    }
```
perf2711, line 770: Should we instead set `x` and/or `y` to 0, and clip what we see?
Same question for the next check about `imageWidth` and `imageHeight`.

Note that this can result in an empty image with dimensions of 0x0.

aga-szczepanska (approved):

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -729,6 +752,54 @@ private DiffingMethodTolerance getDiffingMethodTolerance(CheckOptions checkOptio
...
+
+    return null;
+  }
+
+  private byte[] cropImage(
+      byte[] imageData, String imageFormat, int x, int y, int width, int height) {
+    // We can check that before loading the image
+    if (x < 0 || y < 0) {
+      throw new VisualApiException("Clipped element is not visible.");
```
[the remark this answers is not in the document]
omacranger: I'm not certain for all instances of it in this SDK. Perhaps just to reduce the number of handled exceptions?

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
omacranger: In our other SDKs and the backend we don't throw if the clip element cannot be found, we just silently continue. It's also possible to have a `x` and `y` less than zero but the element still be visible on the page, no? Such as being slightly off screen? Though, it also looks like we scroll the page to the element prior to clipping when not in FPS to get it as much in screen as possible. May want to do that here prior to getting your scroll offsets.

I don't think we're preemptively handling any cases like this in the other SDKs. If it crops to be really small, well, that is what was requested.

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -729,6 +752,54 @@ private DiffingMethodTolerance getDiffingMethodTolerance(CheckOptions checkOptio
...
+    }
+
+    BufferedImage image = loadImage(imageData);
+
+    int imageWidth = image.getWidth();
+    int imageHeight = image.getHeight();
+    if (x + width > imageWidth || y + height > imageHeight) {
```
omacranger, line 776: I think we can still crop even if it's not fully on the screen (this will be better covered with FPS), but we can just `Math.min(imageWidth, x + width)`. I think this is what we're doing in other places.

perf2711: I updated the code to create a crop using the intersection of the visible viewport and clipped region.

omacranger (approved):

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
omacranger: I think that's fine. IIRC selenium throws an exception with `findElement` / uses it as an assertion so that would throw if it were an invalid element anyway. I think that's fine if it would result in an invalid crop

visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -800,20 +793,59 @@ private byte[] saveImage(BufferedImage image, String imageFormat) {
...
     int scrollX = rawScrollX instanceof Long ? ((Long) rawScrollX).intValue() : 0;
     int scrollY = rawScrollY instanceof Long ? ((Long) rawScrollY).intValue() : 0;
+    int width = rawWidth instanceof Long ? ((Long) rawWidth).intValue() : 0;
+    int height = rawHeight instanceof Long ? ((Long) rawHeight).intValue() : 0;
+
+    return new Rectangle(scrollX, scrollY, height, width);
+  }
+
+  private Rectangle scrollTo(Point point) {
+    driver.executeScript(String.format("window.scrollTo(%d, %d)", point.x, point.y));
```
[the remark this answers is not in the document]
perf2711: Done. I also extracted some code to helpers.

PR #233, opened 2025-05-09
[Java] Integration tests

omacranger:
## Description
Adds integration tests to cover the various features we expect the visual SDK to either handle or pass to the backend to handle. Uses the returned visual check snapshot ID to re-query the snapshot and assert that the result matches on following runs (including all properties, ignore region positions, etc). This also exposes a few pieces of functionality to the client to allow others to also make use of some of the features. 

- `sauceVisualCheck` now returns the ID of the snapshot taken, so it can be queried and asserted on if needed
- `visualApi.execute` is now exposed, allowing GraphQL calls through our client so consumers can make additional queries without needing spin up the same configuration

## Types of Changes

- Configuration change
- Refactor/improvements


visual-java/src/test/java/com/saucelabs/visual/integration/ClippingIT.java
```diff
@@ -0,0 +1,59 @@
...
+        sauceVisualCheck(
+            "Scroll to element before clipping",
+            new CheckOptions.Builder()
+                .withClipElement(
+                    driver.findElement(By.cssSelector(".inventory_item:nth-child(21)")))
+                .build());
+    String result = getSnapshotResult(id);
+    expect.toMatchSnapshot(result);
+  }
+}
```
perf2711, line 59: Can we also test elements that are larger than the viewport? E.g. snapping the whole `body`?

visual-java/src/test/java/com/saucelabs/visual/integration/IntegrationBase.java
```diff
@@ -0,0 +1,155 @@
...
+  @BeforeAll
+  public static void setUp() throws MalformedURLException {
+    MutableCapabilities caps = new MutableCapabilities();
+    caps.setCapability("browserName", "chrome");
+    caps.setCapability("browserVersion", "latest");
+    caps.setCapability("platformName", "Windows 10");
+    Map<String, Object> sauceOptions = new HashMap<>();
+    sauceOptions.put("screenResolution", "1600x1200");
+    caps.setCapability("sauce:options", sauceOptions);
+    driver = isLocal ? new FirefoxDriver() : new RemoteWebDriver(getDriverUrl(), caps);
```
perf2711, line 61: For `FirefoxDriver`, we should also make sure to set the screen resolution to 1600x1200.

omacranger: Screen resolution is only for the VM (Sauce) side, however, it does make sense to also set the window size which can be done for both so I shall add that

visual-java/src/test/java/com/saucelabs/visual/integration/ClippingIT.java
omacranger: This should already be done with the `.inventory_list` selector on the `InventoryLongPage`.

visual-java/src/test/java/com/saucelabs/visual/integration/IntegrationBase.java
omacranger: Added setting of window size across both types. This will probably need some additional work to run local & remote tests side by side (since they would share snapshots now which would conflict and we'll need browsers on our Actions image), but this would atleast get us Sauce coverage for right now

perf2711: Why did you set it to something else than for Sauce?

perf2711:
Can you also update the PR's description?

perf2711 (approved):

visual-java/src/test/java/com/saucelabs/visual/integration/IntegrationBase.java
omacranger: The `screenResolution` value is a sauce-specific value for the resolution of the entire _screen_, where setting the size of the window through selenium affects both local and remote executions so we can match them. It also generally has to be a smaller value than `screenResolution` since the window size needs to exclude any sort of OS toolbars.

PR #234, opened 2025-05-13
[no-ticket] [Java/Local] visual-java: do not include ignore regions outside of image

perf2711:
# [Java/Local] visual-java: do not include ignore regions outside of image

## Description
Skips ignore regions not intersecting with the viewport. The viewport may be a clipped element.

## Types of Changes

### DIFF
--- a/visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
+++ b/visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
@@ -659,11 +659,20 @@ public class VisualApi {
       }
     }
 
+    List<RegionIn> visibleIgnoreRegions = new ArrayList<>();
     for (RegionIn region : ignoreRegions) {
-      Point newPoint =
-          CartesianHelpers.relativeTo(viewport.getPoint(), new Point(region.getX(), region.getY()));
-      region.setX(newPoint.x);
-      region.setY(newPoint.y);
+      Rectangle regionRect =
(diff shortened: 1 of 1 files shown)


visual-java/src/main/java/com/saucelabs/visual/VisualApi.java
```diff
@@ -659,11 +659,20 @@ private void sauceVisualCheckLocal(String snapshotName, CheckOptions options) {
...
     }
 
+    List<RegionIn> visibleIgnoreRegions = new ArrayList<>();
     for (RegionIn region : ignoreRegions) {
-      Point newPoint =
-          CartesianHelpers.relativeTo(viewport.getPoint(), new Point(region.getX(), region.getY()));
-      region.setX(newPoint.x);
-      region.setY(newPoint.y);
+      Rectangle regionRect =
+          new Rectangle(region.getX(), region.getY(), region.getHeight(), region.getWidth());
```
omacranger, line 665: Shouldn't this account for the relative changes first ('before' of this diff) prior to checking if they're in the viewport?

Pseudocode: 

```
// mutate point based on viewport changes

// If intersects with viewport, then add as a visible region
```

perf2711: No, as `viewport` and `regionRect` are absolute here (relative to the `(0, 0)` point).

We modify the points to be relative to the image itself (which is relative to `(viewport.x, viewport.y)`) afterwards.

omacranger (approved):

PR #241, opened 2025-06-01
[C#] fix logic so the default region can be accessible

titusfortner:
# One-line summary

Default region inaccessible with existing logic

## Description
Get an error if Region is not provided

## Types of Changes


titusfortner:
I added tests, which I should have done the first time

titusfortner:
never mind, getting the tests right means mucking with environment variable wrapping things that are way out of scope.

PR #244, opened 2025-06-05
[INT-324] Allow to pass WebDriver to the VisualCheck method

perf2711:
# Allow to pass `WebDriver` to the `VisualCheck` method

> Issue : INT-324

## Description
If you want to use `VisualClient` concurrently with multiple drivers, there's no current way to do so.

One could create multiple `VisualClient`s, but this approach has an issue with closing the clients. Every client by design will have the same build, and closing any one of them will cause others to start failing, as they will submit their snapshots to an already closed build.

### Solution
To allow for using multiple drivers with a single build, you can create `VIsualClient` without specifying a `WebDriver` instance. You specify it then in `VisualCheck` method. Using the `VisualCheck` method without specifying a `WebDriver` instance will throw an exception if the `VisualClient` instance was initialized without a driver.

The `VisualClient` instance has to be created once and disposed once per concurrent test. Using multiple `VisualClient` with multiple builds is currently not possible and changing that would be a breaking change. One solution is to potentially add an option like `DoNotReuseBuild`, but that is a topic for a different pull request.

## Types of Changes

### DIFF
--- a/visual-dotnet/SauceLabs.Visual.IntegrationTests/LoginPageConcurrent.cs
+++ b/visual-dotnet/SauceLabs.Visual.IntegrationTests/LoginPageConcurrent.cs
@@ -14,7 +14,7 @@ public class LoginPageConcurrent
     [OneTimeSetUp]
     public async Task Setup()
     {
-        VisualClient = await VisualClient.Create(Utils.GetSauceRegion());
+        VisualClient = await VisualClient.Create();
         TestContext.Progress.WriteLine($"Build: {VisualClient.Build.Url}");
     }
 
--- a/visual-dotnet/SauceLabs.Visual.IntegrationTests/LoginPageConcurrent.cs
+++ b/visual-dotnet/SauceLabs.Visual.IntegrationTests/LoginPageConcurrent.cs
@@ -1,0 +1,78 @@
+using System;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using OpenQA.Selenium.Remote;
+
+namespace SauceLabs.Visual.IntegrationTests;
+
+[Parallelizable(ParallelScope.Children)]
+public class LoginPageConcurrent
+{
+    private VisualClient? VisualClient { get; set; }
+
+    [OneTimeSetUp]
+    public async Task Setup()
+    {
+        VisualClient = await VisualClient.Create(Region.UsWest1);
+        TestContext.Progress.WriteLine($"Build: {VisualClient.Build.Url}");
+    }
+
+    [IntegrationTest]
+    [Test]
+    public async Task LoginPage_ShouldOpen1()
+    {
+        await UsingDriver(async (driver) =>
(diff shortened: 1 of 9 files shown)


konraddysput (changes requested):

visual-dotnet/SauceLabs.Visual.IntegrationTests/LoginPageConcurrent.cs
```diff
@@ -0,0 +1,78 @@
...
+
+[Parallelizable(ParallelScope.Children)]
+public class LoginPageConcurrent
+{
+    private VisualClient? VisualClient { get; set; }
+
+    [OneTimeSetUp]
+    public async Task Setup()
+    {
+        VisualClient = await VisualClient.Create(Region.UsWest1);
```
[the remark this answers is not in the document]
perf2711: Updated

visual-dotnet/SauceLabs.Visual/BuildFactory.cs
```diff
@@ -90,20 +90,13 @@ internal static async Task CloseBuilds()
...
-                var build = (await api.Build(buildId)).EnsureValidResponse().Result;
-                if (build == null)
-                {
-                    throw new VisualClientException($@"build {buildId} was not found");
-                }
-
-                return new VisualBuild(build.Id, build.Url, build.Mode);
-            }
-            catch (VisualClientException)
+            var build = (await api.Build(buildId)).EnsureValidResponse().Result;
```
konraddysput, line 93: We still need to have a try/catch block here because GraphQLResponseExtension in the EnsureValidResponse method throws it - lack of the `buildId` can be problematic there.

visual-dotnet/SauceLabs.Visual.IntegrationTests/LoginPageConcurrent.cs
```diff
@@ -0,0 +1,76 @@
...
+[Parallelizable(ParallelScope.Children)]
+[IntegrationTest(skipIfSingle: true)]
+public class LoginPageConcurrent
+{
+    private VisualClient? VisualClient { get; set; }
+
+    [OneTimeSetUp]
+    public async Task Setup()
+    {
+        VisualClient = await VisualClient.Create(Utils.GetSauceRegion());
```
konraddysput, line 17:
```suggestion
        VisualClient = await VisualClient.Create();
```

what do you think about using a default constructor?

konraddysput (commented):
Are you sure droping try/catches is a good idea?

visual-dotnet/SauceLabs.Visual.IntegrationTests/LoginPageConcurrent.cs
perf2711: Updated.

konraddysput (approved):

PR #260, opened 2026-06-08
Biq 1041 cypress 15 support

paweltomaszewskisaucelabs:
# One-line summary
Add support for cypress 14 and 15.

## Description
To add Cypress version 15, support for Node 16 and 18 needs to be dropped.

## Types of Changes

- Configuration change

## Testing strategy

e2e and manual testing

## Deployment Notes

These should highlight any db migrations, feature toggles, etc.

### DIFF
--- a/visual-js/.changeset/nice-fishes-obey.md
+++ b/visual-js/.changeset/nice-fishes-obey.md
@@ -1,5 +1,0 @@
----
-"@saucelabs/cypress-visual-plugin": minor
----
-
-bump node, cypress versions
--- a/visual-js/.changeset/nice-fishes-obey.md
+++ b/visual-js/.changeset/nice-fishes-obey.md
@@ -1,0 +1,5 @@
+---
+"@saucelabs/cypress-visual-plugin": minor
+---
+
+bump node, cypress versions
--- a/visual-js/visual-cypress/package.json
+++ b/visual-js/visual-cypress/package.json
@@ -12,7 +12,7 @@
     "README.md"
   ],
   "engines": {
-    "node": "^16.13 || >=18"
+    "node": "^20 || ^22 || >=24"
   },
   "typeScriptVersion": "5.1.6",
   "keywords": [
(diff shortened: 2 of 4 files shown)


visual-js/.changeset/nice-fishes-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+"@saucelabs/cypress-visual-plugin": minor
```
bogdan-karmyzow: Is dropping 16 and 18 node versions considered a minor version bump?

visual-js/visual-cypress/package.json
```diff
@@ -61,7 +61,6 @@
   },
   "devDependencies": {
     "@tsconfig/node18": "^2.0.1",
-    "@types/cypress": "^1.1.3",
```
bogdan-karmyzow, line 64: Why did we remove the cypress types here?

paweltomaszewskisaucelabs: Why should we keep it? 
https://www.npmjs.com/package/@types/cypress

visual-js/.changeset/nice-fishes-obey.md
paweltomaszewskisaucelabs: yes it is when bumping to the "0.11.0" version

visual-js/visual-cypress/package.json
```diff
@@ -12,7 +12,7 @@
     "README.md"
   ],
   "engines": {
-    "node": "^16.13 || >=18"
```
konraddysput, line 15: why do we need to drop the node version?

visual-js/visual-cypress/package.json
bogdan-karmyzow: So the types were not used before -- we can safely delete them?

konraddysput (approved):

visual-js/visual-cypress/package.json
paweltomaszewskisaucelabs: yes

PR #261, opened 2026-07-02
feat: Pinning the version of kotlin stdlib to 2.4.0

bogdan-karmyzow:
# One-line summary

> Issue : BIQ-1139

## Description

In this PR I am bumping a version of kotlin-stdlib. This package is not used directly but is being pulled in by other packages.
So I pinned the version of the kotlin-stdlib to be 2.4.0

## Types of Changes

- Configuration change

## Testing strategy

Run unit tests form the Iris/e2e repo while using the latest espress sdk version

## Deployment Notes

We will trigger the release pipeline

### DIFF
--- a/visual-espresso/visual/build.gradle
+++ b/visual-espresso/visual/build.gradle
@@ -55,6 +55,8 @@ String getProperty(String name, String defaultValue) {
 }
 
 dependencies {
+    implementation platform(libs.kotlin.bom)
+
     implementation libs.appcompat
     implementation libs.material
     implementation libs.junit
--- a/visual-espresso/visual/build.gradle
+++ b/visual-espresso/visual/build.gradle
@@ -56,7 +56,6 @@ String getProperty(String name, String defaultValue) {
 
 dependencies {
     implementation platform(libs.kotlin.bom)
-
     implementation libs.appcompat
     implementation libs.material
     implementation libs.junit
--- a/visual-espresso/gradle/libs.versions.toml
+++ b/visual-espresso/gradle/libs.versions.toml
@@ -1,6 +1,7 @@
 [versions]
 agp = "8.5.2"
 apolloGraphQL = "3.8.5"
+kotlin = "2.4.0"
 jsoup = "1.18.1"
 junit = "4.13.2"
 junitVersion = "1.2.1"
@@ -11,6 +12,7 @@ uiautomator = "2.2.0"
 vanniktechMavenPublish = "0.30.0"
 
 [libraries]
+kotlin-bom = { group = "org.jetbrains.kotlin", name = "kotlin-bom", version.ref = "kotlin" }
 jsoup = { group = "org.jsoup", name = "jsoup", version.ref = "jsoup" }
 junit = { group = "junit", name = "junit", version.ref = "junit" }
 ext-junit = { group = "androidx.test.ext", name = "junit", version.ref = "junitVersion" }


visual-espresso/visual/build.gradle
```diff
@@ -55,6 +55,8 @@ String getProperty(String name, String defaultValue) {
 }
 
 dependencies {
+    implementation platform(libs.kotlin.bom)
+
```
melekr: extra space here

melekr (commented):
Before merging, can we add evidence that the published SDK metadata also causes downstream consumers to resolve `org.jetbrains.kotlin:kotlin-stdlib `to `2.4.0 or newer`?

melekr:
The current CI only runs `./gradlew build`, but the full the release path publishes to maven. 
A `dependencyInsight` on `releaseRuntimeClasspath`, plus `publishToMavenLocal` consumer check would make the changes verifiable;

> ./gradlew :visual:dependencyInsight \
  --dependency kotlin-stdlib \
  --configuration releaseRuntimeClasspath
  
  

> ./gradlew :visual:publishToMavenLocal

bogdan-karmyzow:
@melekr I ran both commands and they pass:

- `./gradlew :visual:dependencyInsight --dependency kotlin-stdlib --configuration releaseRuntimeClasspath` now shows `kotlin-stdlib` forced to 2.4.0 (beating the flagged 1.8.22).
- `./gradlew :visual:publishToMavenLocal` also publishes cleanly.

But the catch is that both of those only prove the bump works on the SDK's own build. When I went one step further and actually consumed the published SDK from a throwaway test project (as you asked) the consumer still resolves `kotlin-stdlib` to 1.9.21 (a 1.x, still below the 2.1.0 the advisory requires), and `kotlin-stdlib-jdk7/jdk8` stay at 1.8.22.

So the version force doesn't propagate through the published AAR to downstream consumers - a consumer ends up with whatever its own dependency graph resolves (1.9.21 in my test; likely 1.8.22 in a graph like the customer's). Only `kotlin-stdlib` is affected this way; every other dep propagates fine (as I have researched it's a known quirk of forcing a transitive JVM stdlib through an Android AAR).

So this PR alone won't clear the scan on the consumer side. Two ways forward:

1. Bump the deps that actually drag in old Kotlin - mainly Apollo 3.8.5 → 4.x (built on Kotlin 2.x). Those edges do propagate, so newer Kotlin would reach consumers (bigger change, with breaking-API risk, and needs testing).
2. it can be fixed on a consumer-side with a ~3-line override in the consumer's build. Reliable and immediate, but puts the work on the consumer instead of shipping a clean SDK.

visual-espresso/visual/build.gradle
bogdan-karmyzow: Fixed

melekr:
@bogdan-karmyzow makes sense, we usually do both;
- Convert this or supersede it with an sdk-side fix (since Apollo migration is non trivial)
- Meanwhile, document the consumer-side gradle override as the immediate workaround to communicate/unblock customers, up to @konraddysput & CS discretion.

konraddysput:
Lets make sure its added to sauce docs

bogdan-karmyzow:
Here is [the PR](https://github.com/saucelabs/sauce-docs/pull/3582) to add a documentation for the consumer-side gradle override. @melekr could you please check if the wording looks correct and so on?

And here is [a ticket](https://saucedev.atlassian.net/browse/BIQ-1147?atlOrigin=eyJpIjoiN2JlY2VmMGRmZGE2NGQ5N2FjODg4Njg1YjJhYTgyNjIiLCJwIjoiaiJ9) to bump the apollo version to 5.0.1 (latest) which will 100% bump the version of transitive kotlin-stdlib to 2.1.0 @konraddysput

bogdan-karmyzow:
@melekr @konraddysput pinging for visibility

bogdan-karmyzow:
@melekr are we good to merge this MR? Do we need to do anything after the merge to make a release?

melekr (approved):
@bogdan-karmyzow Good to go to immediately unblock with the documented workaround. Let's work on shipping the apollo migration asap. cc: @konraddysput

PR #264, opened 2026-07-16
[Espresso SDK] Bumped version of apollo to 5.0.1

bogdan-karmyzow:
# [visual-espresso] Migrate Apollo 3.8.5 → 5.0.1 to clear kotlin-stdlib CVE

> Issue: BIQ-1147

## Description

In this PR I am upgrading Apollo to 5.0.1 so downstream consumers naturally resolve kotlin-stdlib ≥ 2.1.0:

As a bonus, removing the kotlin 2.4.0 pin clears the non-fatal `Module was compiled with an incompatible version of Kotlin` lint errors seen in CI.

## Types of Changes

- Configuration change
- Refactor/improvements

## Testing strategy

- `./gradlew :visual:build` (incl. lint) passes with Gradle 8.7 / AGP 8.6.1 / JDK 17
- `publishToMavenLocal` + throwaway consumer: `dependencyInsight` shows every `kotlin-stdlib`
  edge resolving to 2.2.21 (≥ 2.1.0 → Snyk issue cleared); legacy `kotlin-stdlib-jdk7/8:1.8.0`
  entries are JetBrains' own alignment constraints with no known vulnerabilities
- Sample consumer app compiled against the published AAR (`assembleDebugAndroidTest`),
  exercising the exposed Apollo API surface (`GraphQLClient(ApolloClient)`, `executeMutation`)
- Device `SmokeTest` on an android-34 emulator: 2/2 passed, snapshots visible in Sauce Visual
  (project "Espresso SDK", build "Smoke test")

## Deployment Notes

- No new secrets or infra changes
- Next release publishes an AAR without RxJava3 and with Apollo 5.0.1; consumers resolve kotlin-stdlib ≥ 2.1.0 naturally

### DIFF
--- a/visual-espresso/visual/src/main/java/com/saucelabs/visual/graphql/GraphQLClient.java
+++ b/visual-espresso/visual/src/main/java/com/saucelabs/visual/graphql/GraphQLClient.java
@@ -10,6 +10,7 @@ import com.apollographql.apollo.api.Error;
 import com.apollographql.apollo.api.Mutation;
 import com.apollographql.apollo.api.Operation;
 import com.apollographql.apollo.api.Query;
+import com.apollographql.apollo.exception.ApolloException;
 import com.saucelabs.visual.BuildConfig;
 import com.saucelabs.visual.DataCenter;
 import com.saucelabs.visual.exception.VisualApiException;
@@ -59,7 +60,10 @@ public class GraphQLClient {
                     (scope, continuation) -> call.execute(continuation));
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
-            throw new VisualApiException("Interrupted while executing GraphQL call");
+            throw new VisualApiException("Interrupted while executing GraphQL call", e);
+        } catch (ApolloException e) {
+            String message = e.getMessage() != null ? e.getMessage() : "GraphQL call failed";
+            throw new VisualApiException(message, e);
         }
     }
 
@@ -76,7 +80,10 @@ public class GraphQLClient {
             }
             throw new VisualApiException(message.toString());
         } else if (response.exception != null) {
-            throw new VisualApiException(response.exception.getMessage());
+            String message = response.exception.getMessage() != null
+                    ? response.exception.getMessage()
+                    : "GraphQL request failed";
+            throw new VisualApiException(message, response.exception);
         } else {
             throw new VisualApiException("Unexpected GraphQL error");
         }
--- a/visual-espresso/visual/src/main/java/com/saucelabs/visual/graphql/GraphQLClient.java
+++ b/visual-espresso/visual/src/main/java/com/saucelabs/visual/graphql/GraphQLClient.java
@@ -3,20 +3,19 @@ package com.saucelabs.visual.graphql;
 import android.text.TextUtils;
 import android.util.Base64;
 
-import com.apollographql.apollo3.ApolloCall;
-import com.apollographql.apollo3.ApolloClient;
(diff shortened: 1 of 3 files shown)


visual-espresso/visual/src/main/java/com/saucelabs/visual/graphql/GraphQLClient.java
```diff
@@ -79,6 +75,8 @@ private <D extends Operation.Data> D handleResponse(ApolloResponse<D> response)
                 message.append(error.getMessage());
             }
             throw new VisualApiException(message.toString());
+        } else if (response.exception != null) {
```
melekr, line 82: Could we preserve the original Apollo exception as the cause here? The current code drops the exception type and stack trace and getMessage() may be null.

`VisualApiException` already supports wrapping a cause, so this would retain useful diagnostic information:
```suggestion
        } else if (response.exception != null) {
    String message = response.exception.getMessage() != null
            ? response.exception.getMessage()
            : "GraphQL request failed";
    throw new VisualApiException(message, response.exception);
}
```

visual-espresso/visual/src/main/java/com/saucelabs/visual/graphql/GraphQLClient.java
```diff
@@ -45,25 +44,22 @@ public GraphQLClient(ApolloClient apolloClient) {
...
         try {
-            ApolloResponse<D> response = single.blockingGet();
-            return handleResponse(response);
-        } catch (ApolloHttpException e) {
-            throw new VisualApiException(e.getMessage());
+            return (ApolloResponse<D>) BuildersKt.runBlocking(
+                    EmptyCoroutineContext.INSTANCE,
+                    (scope, continuation) -> call.execute(continuation));
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
```
melekr, line 62: same here, we can wrap Apollo failures in VisualApiException while preserving the cause;

```suggestion
            Thread.currentThread().interrupt();
            throw new VisualApiException("Interrupted while executing GraphQL call", e);
         } catch (ApolloException e) {
            String message = e.getMessage() != null
                      ? e.getMessage()
                      : "GraphQL call failed";
             throw new VisualApiException(message, e);
         }
```

melekr (commented):
Thanks for taking this migration on @bogdan-karmyzow . 

One compatibility concern: Since `GraphQLClient` exposes Apollo types in its public constructors/methods, the change from `com.apollographql.apollo3.* `to `com.apollographql.apollo.*` can make the release source/binary breaking for consumers using those APIs. 

- Could we update the PR + add a short migration note?
- If not already tested, could we also compile with a sample app;
   `./gradlew :visual:publishToMavenLocal`
    `./gradlew -p <sample-project> assembleDebugAndroidTest`

visual-espresso/visual/src/main/java/com/saucelabs/visual/graphql/GraphQLClient.java
bogdan-karmyzow: Thanks, applied

visual-espresso/visual/src/main/java/com/saucelabs/visual/graphql/GraphQLClient.java
bogdan-karmyzow: Done

bogdan-karmyzow:
Thanks @melekr for the review, I have applied your comments

melekr (approved):

